PR #3, opened 2024-11-27
Include content formula in `formulasInComponent`

erik-beus:
The `content` formula was not visited in the `formulasInComponent` function in the `ToddleComponent` class. This meant that formulas were reported as unused, even when they were used. See reported issue here https://discord.com/channels/972416966683926538/1311015179798184021/1311015179798184021


jacob-kofoed (approved):
I really hope this is the last missing check this time 😱

PR #4, opened 2024-11-27
Update types for ToddleComponent and ToddleFormula

erik-beus:
We were not visiting global formulas when we were looking for formula references. This fixes that by adding new parameters where necessary to allow the different classes/functions to visit global formulas.

Please see reported issue here: https://discord.com/channels/972416966683926538/1310985625205473421/1310985625205473421
and the test project here: https://toddle.dev/projects/broken_global_formulas_project/branches/main/components/HomePage?leftpanel=design&canvas-width=800&canvas-height=925&rightpanel=attributes&selection=nodes.Uh8K4O215YFRyrAjH6eKN

### DIFF
--- a/packages/core/src/component/ToddleComponent.ts
+++ b/packages/core/src/component/ToddleComponent.ts
@@ -103,189 +103,188 @@ export class ToddleComponent<Handler> {
     ): Generator<[(string | number)[], Formula]> {
       switch (node.type) {
         case 'text':
-          yield* getFormulasInFormula(node.condition, globalFormulas, [
-            ...path,
-            'condition',
-          ])
-          yield* getFormulasInFormula(node.repeat, globalFormulas, [
-            ...path,
-            'repeat',
-          ])
-          yield* getFormulasInFormula(node.repeatKey, globalFormulas, [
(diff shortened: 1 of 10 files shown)


jacob-kofoed (approved):
Looks good. Actually a pretty small change, but happens to be the function we call the most times in our code 😄 

You should set up a test project with a global formula calling itself to see if everything dies. Perhaps before merging.

packages/core/src/component/ToddleComponent.ts
```diff
@@ -251,17 +294,31 @@ export class ToddleComponent {
...
         'onAttributeChange',
         'actions',
         actionKey,
       ])
     }
+    // Visit all component formulas in case they reference other formulas
+    for (const [key, componentFormula] of Object.entries(
+      this.component.formulas ?? {},
+    )) {
+      yield* getFormulasInFormula(componentFormula.formula, globalFormulas, [
```
erik-beus, line 316: FYI we also didn't do this before. Component formulas could potentially reference other global formulas

jacob-kofoed (dismissed):

PR #15, opened 2024-12-03
Add FormulaContext logic in SSR package

erik-beus:
- Add getPageFormulaContext function in ssr package
- Improve types of FormulaContext.Data in core package
- Add getRequestCookies function in ssr package
- Add cookie + xss dependencies in ssr package
- Remove index files and related exports in ssr package


packages/ssr/package.json
```diff
@@ -17,6 +17,8 @@
   "main": "dist/index.js",
   "types": "dist/index.d.ts",
   "dependencies": {
-    "@toddledev/core": "workspace:*"
+    "@toddledev/core": "workspace:*",
+    "cookie": "1.0.2",
```
erik-beus, line 21: This is a newer version that we're currently using internally (´0.5.0´)

packages/ssr/src/rendering/cookies.ts
```diff
@@ -0,0 +1,10 @@
+import { isDefined } from '@toddledev/core/dist/utils/util'
+import { parse } from 'cookie'
+
+export const getRequestCookies = (req: Request) =>
+  Object.fromEntries(
+    Object.entries(parse(req.headers.get('cookie') ?? '')).filter(
```
erik-beus, line 6: `parse` returns `Record<string, string | null>`
Therefore, I filter out `nullish` values

examples/hono/src/routes/toddlePage.ts
```diff
@@ -19,16 +21,25 @@ export const toddlePage = async (c: Context<HonoEnv>) => {
...
       <html lang="${language}">
         <head>
           <meta charset="UTF-8" />
+          <title>
+            ${applyFormula(page.route.info?.title?.formula, formulaContext) ??
+            page.name}
+          </title>
         </head>
         <body>
           <h1>${page.name}</h1>
```
jacob-kofoed, line 45: Should render full app here please 🙅

packages/runtime/src/editor-preview.main.ts
```diff
@@ -462,16 +465,23 @@ export const createRoot = (
...
-              Location: component?.page
-                ? {
-                    ...data.Location,
-                    query: attrs,
-                  }
-                : data.Location,
-              Props: attrs ?? {},
-            }))
+            dataSignal.update((data) => {
+              // We should figure out if "Props" is used anywhere and get rid of it if it's not
```
jacob-kofoed, line 469: Good question, maybe @cullophid knows the history of this and if it's used?

packages/ssr/package.json
```diff
@@ -1,14 +1,14 @@
 {
   "name": "@toddledev/ssr",
-  "version": "0.0.2-alpha.3",
+  "version": "0.0.2-alpha.4",
```
jacob-kofoed, line 3: Have we discussed the option to run with a single version across all packages instead of individually? I think this is best for consumers, but we will have a tough time keeping track of which versions are compatible with each other.

jacob-kofoed (approved):

packages/ssr/package.json
erik-beus: We should look into that eventually yes. Right now we only have 1 real consumer though, so I think it's fine to postpone to keep things simple and to avoid too much manual work when doing changes.

PR #16, opened 2024-12-03
Limit function parameters to 3

erik-beus:
Let's help ourselves by limiting the number of parameters to 3. This will make our functions easier to understand and maintain.


ChrEsb (approved):
Also good one

PR #22, opened 2024-12-06
Add rule for dl->dd|dt|script|template relation

erik-beus:
I found another element where the direct child rule is relevant (although I've never heard of the `dl`, `dt` or `dd` element 😂 )


jacob-kofoed (approved):
Dunno how we survived without this rule this far 😄

PR #24, opened 2024-12-10
Move core formulas from global to FormulaContext

erik-beus:
We want to use globalThis/window less, so we need to move core+custom formulas to FormulaContext.
Before this is ready to merge, we must find a solution for using `isEqual` in our lib formulas (that currently rely on globalThis/window).


jacob-kofoed (approved):
Tested and works

PR #27, opened 2024-12-13
Add rule for unknown events

erik-beus:
See suggestion here https://discord.com/channels/972416966683926538/1316702664267923507


jacob-kofoed (approved):
Looks good. Surprised we didn't have this yet, so useful!

PR #28, opened 2024-12-13
Add rule to detect unused events

erik-beus:
- Add new rule for unused events
- Adjust types with an explicit type for component events
- Bump version in package.json of core + search

Related to https://github.com/toddledev/toddle/pull/27


jacob-kofoed (approved):
Awesome

PR #29, opened 2024-12-16
Only pass numbers to value property of <progress>

erik-beus:
Tom reported an issue, where an `HTMLProgressElement` would fail to render + break other elements, if the passed value attribute was not a finite number (or a string that can be casted into a finite number).

### DIFF
--- a/packages/runtime/src/utils/setAttribute.ts
+++ b/packages/runtime/src/utils/setAttribute.ts
@@ -1,4 +1,4 @@
-import { toBoolean } from '@toddledev/core/dist/utils/util'
+import { isDefined, toBoolean } from '@toddledev/core/dist/utils/util'
 
 /**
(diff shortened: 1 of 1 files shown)


packages/runtime/src/utils/setAttribute.ts
```diff
@@ -18,9 +18,16 @@ export function setAttribute(
       }
       break
     case 'value':
-    case 'type':
-      ;(elem as any)[attr] = toBoolean(value) ? String(value) : undefined
+    case 'type': {
+      let val = value
+      if (elem instanceof HTMLProgressElement) {
```
jacob-kofoed, line 23: Maybe a super short comment here for the future, or link to the original issue

erik-beus: Done

PR #32, opened 2024-12-17
Run api actions after all apis are initialised

ChrEsb:


packages/runtime/src/api/createAPIv2.ts
```diff
@@ -41,7 +41,12 @@ import { ComponentContext } from '../types'
 export function createAPI(
   apiRequest: ApiRequest,
   ctx: ComponentContext,
-): { fetch: Function; update: Function; destroy: Function } {
+): {
+  fetch: Function
```
erik-beus, line 45: Perhaps we could add a short description of why we have multiple functions and in which order they're supposed to be called?

packages/runtime/src/api/createAPIv2.ts
```diff
@@ -854,6 +885,25 @@ export function createAPI(
...
     },
+    triggerActions: () => {
+      const apiData = ctx.dataSignal.get().Apis?.[api.name]
+      if (
+        apiData === undefined ||
+        (apiData.data === null && apiData.error === null)
+      ) {
+        return
+      }
+      if (apiData.error) {
```
erik-beus, line 896: Does this take the error formula into account?

packages/runtime/src/components/createComponent.ts
```diff
@@ -159,6 +159,11 @@ export function createComponent({
       })
     }
   })
+  Object.entries(apis)
+    .filter(([_, api]) => api.triggerActions !== undefined)
```
erik-beus, line 163: Why/when would `triggerActions` be `undefined`? To avoid the optional chaining on line 165, you could do something like this perhaps. Also I suppose there's no reason to use `Object.entries` if you only use the values:
```
Object.values(apis)
	.filter((api): api is RequireFields<ContextApi, 'triggerActions'> => api.triggerActions !== undefined)
```

packages/runtime/src/types.d.ts
```diff
@@ -54,10 +54,7 @@ export interface ComponentContext {
   isRootComponent: boolean
   dataSignal: Signal<ComponentData>
   triggerEvent: (event: string, data: unknown) => void
-  apis: Record<
-    string,
-    { fetch: Function; destroy: Function; update?: Function }
-  >
+  apis: ContextApi
```
erik-beus, line 57: Consider changing this to
```suggestion
  apis: Record<string, ContextApi>
```

packages/runtime/src/types.d.ts
```diff
@@ -78,6 +75,16 @@ export interface ComponentContext {
...
 
+export type ContextApi = Record<
+  string,
+  {
+    fetch: Function
+    destroy: Function
+    update?: Function
+    triggerActions?: Function
+  }
+>
```
erik-beus, line 86: This would make it easier to work with the type 🤞 
```suggestion
export interface ContextApi = {
    fetch: Function
    destroy: Function
    update?: Function
    triggerActions?: Function
  }
```

erik-beus (changes requested):
Please see questions/suggestions :)

packages/runtime/src/api/createAPIv2.ts
ChrEsb: Currently this works by applying the actions to the currently set data. Before the data is set on the signal the `isError` formula is applied. So I believe it is safe to assume that `error` is already correct. It is the same logic used when determining if its a success or error when hydrating from ssr apis.

erik-beus (approved):

PR #37, opened 2024-12-20
ci/npm publish

erik-beus:
Allow publishing to npm from Github Actions


packages/ssr/package.json
```diff
@@ -1,9 +1,8 @@
 {
   "name": "@toddledev/ssr",
-  "version": "0.0.2-alpha.6",
+  "version": "0.0.2-alpha.13",
```
erik-beus, line 3: This was already published to npm

bunfig.toml
```diff
@@ -0,0 +1,2 @@
+[install]
+saveTextLockfile = true
```
jacob-kofoed, line 2: Not sure what the consequences would be if someone is on bun 1.1.38 or older 🤔

jacob-kofoed (approved):
I didn't test publish workflow, but looks correct. Do we need to write in README how publishing works? Where to bump versions etc.

erik-beus:
> I didn't test publish workflow, but looks correct. Do we need to write in README how publishing works? Where to bump versions etc.

Good point. I'll get that added (in a followup PR)

PR #39, opened 2024-12-20
Use canvas highlight instead of custom container outline

jacob-kofoed:
Andreas suggested using the canvas element hover instead of a custom outline—this way the drop-target is also highlighted in the tree view.

### DIFF
--- a/packages/runtime/src/editor-preview.main.ts
+++ b/packages/runtime/src/editor-preview.main.ts
@@ -841,6 +841,12 @@ export const createRoot = (
               }
               break
           }
+          if (message.data.canceled) {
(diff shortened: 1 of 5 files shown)


packages/runtime/src/editor-preview.main.ts
```diff
@@ -841,6 +841,12 @@ export const createRoot = (
               }
               break
           }
+          if (message.data.canceled) {
```
jacob-kofoed, line 844: This is only conditionally on cancel (ESC during drag) to make the drop feel instant instead of waiting for the DOM update.

PR #41, opened 2025-01-08
Add support for rendering head tags during SSR

erik-beus:
- Add support for rendering head tags during SSR
- Allow ordering of head tags using a custom order
- Add simple nanoid function to generate unique ids
- Add font mapping function - this requires a font endpoint in the hono example. This PR should not be merged until that's ready
- Update hono example and provide the project using var over env - see https://hono.dev/docs/api/context#var

### DIFF
--- a/examples/hono/src/routes/toddlePage.ts
+++ b/examples/hono/src/routes/toddlePage.ts
@@ -85,8 +85,8 @@ export const toddlePage = async (c: Context<HonoEnv>) => {
       packages: project.files.packages,
     },
   })
-  const head = renderHeadItems(
-    getHeadItems({
+  const head = renderHeadItems({
+    headItems: getHeadItems({
       url,
       // This refers to the endpoint we created in fontRouter for our proxied stylesheet
       cssBasePath: '/.toddle/fonts/stylesheet/css2',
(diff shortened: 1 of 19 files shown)


examples/hono/src/routes/toddlePage.ts
```diff
@@ -79,44 +81,54 @@ export const toddlePage = async (c: Context<HonoEnv>) => {
...
+      page: component,
+      files: project.files,
+      project: project.project,
+      context: formulaContext,
+      theme,
+    }),
+  )
   const { html: body } = await renderPageBody({
-    component: toddleComponent as any, // TODO: Fix typing
+    component: component,
```
ChrEsb, line 99:
```suggestion
    component,
```

examples/hono/src/routes/toddlePage.ts
```diff
@@ -79,44 +81,54 @@ export const toddlePage = async (c: Context<HonoEnv>) => {
...
-            href="/_static/reset.css"
-          />
+          ${raw(head)}
           <style>
             ${styles}
           </style>
         </head>
         <body>
-          ${raw(body)}
+          <!-- ${c.req.raw.url} -->
```
ChrEsb, line 124: Is this to make some debugging easier?

ChrEsb (changes requested):
Looks good. Just a minor thing and a question

examples/hono/src/routes/toddlePage.ts
erik-beus: It was actually something we rendered internally, but I don't think it adds a lot of value tbh. I'll remove it

ChrEsb (approved):

PR #49, opened 2025-01-14
POC of simple hydration

jacob-kofoed:
Resumability is still a while off. Until then, hydration would fix all visual issues of double-render, except performance (sending data multiple times that resumability handles).

This is a proof of concept and unrealistic to complete before we have some sort of runtime versioning.

### DIFF
--- a/packages/ssr/src/rendering/components.ts
+++ b/packages/ssr/src/rendering/components.ts
@@ -187,11 +187,11 @@ const renderComponent = async ({
             : node.tag
         const nodeClasses = `${classHash} ${classList}`.trim()
         if (!VOID_HTML_ELEMENTS.includes(tag)) {
-          return `<${tag} ${nodeAttrs} data-node-id="${escapeAttrValue(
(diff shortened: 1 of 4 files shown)


packages/ssr/src/rendering/components.ts
```diff
@@ -187,11 +187,11 @@ const renderComponent = async ({
             : node.tag
         const nodeClasses = `${classHash} ${classList}`.trim()
         if (!VOID_HTML_ELEMENTS.includes(tag)) {
-          return `<${tag} ${nodeAttrs} data-node-id="${escapeAttrValue(
+          return `<${tag} ${nodeAttrs} data-node-component="${component.name}" data-node-id="${escapeAttrValue(
```
erik-beus, line 190: We should probably also escape this attribute? 🤔
```suggestion
          return `<${tag} ${nodeAttrs} data-node-component="${escapeAttrValue(component.name)}" data-node-id="${escapeAttrValue(
```

erik-beus (commented):
Looks promising 👀 Lets try it out on a few projects. But yes, we probably can't use this approach until we have versioning 👍

PR #51, opened 2025-01-16
Add rule for detecting nodes with no references

erik-beus:
I plan to add a feature flag which will exclude this rule for all users except for users who opt in. Currently, it's not possible to perform any action based on this rule, but it's an obvious candidate for a rule that could be used to automatically delete nodes that are not referenced by any other nodes.

In the toddle project, this rule creates 39 issues (meaning 39 different components). If we flag each unused node (multiple times per component), this rule would create 63 issues.


jacob-kofoed (approved):
As you say, it is not a fixable issue. To me, this falls into a separate category of corrupted data rather than an issue. Something like this should ultimately be cleaned up just before saving the data. This change is great for visibility around corrupted data and as a backup in the future, but perhaps it is soon time to look into schema-validation again and ways to ensure what we store is valid.

PR #52, opened 2025-01-17
Use a BatchQueue for the component lifecycle callback

jacob-kofoed:
I looked into Spark and the Editor for some performance improvements. Simply registering the ticks via. multiple (481 on spark) took 41ms, and then running each update in a different tick added an additional ~100ms.

On the toddle editor this change seems to reduce startup scripting time from 1653ms -> 1258ms (24% improvement) and should also benefit show/hide and dynamic repeats.

This is not enough on its own to get Spark a green lighthouse score, but it should help a bit.

Pretty average snapshot of render flamegraph. Each of the yellow fields was the allocation (not actually running the callback) to requestAnimationFrame. This is essentially 0ms after this update.

CleanShot 2025-01-17 at 14 04 06@2x

83 microseconds called 418 times is ~40ms

### DIFF
--- a/packages/runtime/src/utils/BatchQueue.ts
+++ b/packages/runtime/src/utils/BatchQueue.ts
@@ -1,0 +1,25 @@
+/**
+ * A helper class to batch multiple callbacks and process them in a single tick.
+ * This is more efficient than processing each callback in a separate tick, as creating a new tick is expensive.
+ * It also allows batching DOM updates, which can help to reduce layout thrashing.
+ */
+export class BatchQueue {
+  private batchQueue = new Set<() => void>()
+  private isProcessing = false
+
+  private processBatch() {
+    if (this.isProcessing) return
+    this.isProcessing = true
+
+    setTimeout(() => {
+      this.batchQueue.forEach((callback) => callback())
+      this.batchQueue.clear()
+      this.isProcessing = false
+    })
+  }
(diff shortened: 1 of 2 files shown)


packages/runtime/src/utils/BatchQueue.ts
```diff
@@ -0,0 +1,25 @@
...
+
+    setTimeout(() => {
+      this.batchQueue.forEach((callback) => callback())
+      this.batchQueue.clear()
+      this.isProcessing = false
+    })
+  }
+
+  public add(callback: () => void) {
+    this.batchQueue.add(callback)
```
erik-beus, line 22: Would it be possible to have a situation where a callback is added while there batch is being processed? Not sure if that could be an issue

erik-beus (commented):
Very interesting! Let's test 👍

packages/runtime/src/utils/BatchQueue.ts
jacob-kofoed: It could be possible, but it would still render after the component setup as it would go to the end of the queue. We could add a form of batch buffer to offset new entries after it has started, but the important part is just that the `onLoad` is deferred to after the rendering, which it would still be even if added during processing 🤔

I will try to set up a specific test-case for this to be sure 👍

PR #53, opened 2025-01-20
Add support for custom routes (rewrites)

erik-beus:
- Update types
- Update hono example
- Add project file that holds a route example
- Add function for finding a matching route with the same approach as for finding a matching page

### DIFF
--- a/packages/ssr/src/rendering/formulaContext.ts
+++ b/packages/ssr/src/rendering/formulaContext.ts
@@ -40,12 +40,6 @@ export const getPageFormulaContext = ({
   const { searchParamsWithDefaults, hash, combinedParams, url } = getParameters(
     { route: component.route, req },
   )
-  const coreFormulas = Object.fromEntries(
-    Object.entries(libFormulas).map(([name, module]) => [
-      '@toddle/' + name,
-      module.default as any,
-    ]),
-  )
   const formulaContext: FormulaContext & { env: ToddleServerEnv } = {
     data: {
       Location: {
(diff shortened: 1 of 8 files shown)


packages/ssr/src/rendering/formulaContext.ts
```diff
@@ -56,10 +59,7 @@ export const getPageFormulaContext = ({
       // Path and query parameters are referenced in a flat structure in formulas
       // hence, we need to merge them. We prefer path parameters over query parameters
       // in case of naming collisions
-      'URL parameters': {
-        ...searchParamsWithDefaults,
-        ...combinedParams,
-      } as Record<string, string>,
+      'URL parameters': getDataUrlParameters({ route: component.route, req }),
```
mikkeltoddle, line 62: It seems you are already getting these from the first `getParameters` call. What is the purpose of calling it again through `getDataUrlParameters` (which basically just calls `getParameters`)?

packages/ssr/src/routing/routing.ts
```diff
@@ -42,6 +80,29 @@ export const matchPageForUrl = ({
...
+      // destination formulas should only have access to URL parameters from
+      // the route's source definition. Not from anything else atm.
+      {
+        data: {
+          'URL parameters': getDataUrlParameters({
+            route: route.source,
+            req,
+          }),
+        },
+      } as any,
```
mikkeltoddle, line 102: Will they also have access to global formulas? 🤔

erik-beus: Yes, I think they should have. It was just a bit tricky to give them access right now, but that could be added later 👍

packages/ssr/src/rendering/formulaContext.ts
erik-beus: It was to ensure we construct them in the same way both here and in `routing.ts`. Perf wise, it shouldn't matter that we call `getParameters` twice I think.

packages/ssr/src/ssr.types.ts
```diff
@@ -57,3 +61,10 @@ export type PluginAction = {
   // exported indicates that an action is exported in a package
   exported?: boolean
 }
+
+export interface Route {
+  source: Pick<PageRoute, 'path' | 'query'>
```
jacob-kofoed, line 66: The types are a little confusing to me. Since we have two types of routes, maybe it is better if they inherit the same base class, instead of one being a subset of the other 🤔

packages/ssr/src/ssr.types.ts
```diff
@@ -44,7 +48,7 @@ export type InstalledPackage = Pick<
   }
 }
 
-export type PluginAction = {
+export interface PluginAction {
```
jacob-kofoed, line 51: Curious, why change to interface. Is this our convention, or specific reason?

erik-beus: I believe the Typescript team recommends interfaces when possible due to better performance 🤷‍♂️ https://github.com/microsoft/TypeScript/wiki/Performance#preferring-interfaces-over-intersections

PR #60, opened 2025-01-23
Unknown event rule should look up pckg components

erik-beus:
The rule would report incorrect results since it wasn't checking the available events in package components.

See reported issue here https://discord.com/channels/972416966683926538/1331948264106491934


marija04 (approved):
Looks good  👌

PR #62, opened 2025-01-27
Initialize variables after context subscription

jacob-kofoed:
This fixes an issue where some users could not use context values in the variables' initial value.

The easy fix was to set the data signal in a few steps, but this could have some overhead as we deep-equal-checked values even before the signal was used. I updated the signal implementation to short circuit if there are no subscribers. This alone seems to have improved performance slightly for many projects. I also changed our signal to use sets instead of arrays to further improve performance when destructing signals (not related and very minor performance improvement).

This should not be breaking in any way, but please test as well 👍 SSR as well

I used this very simple example to test. It should say "value: 5": https://start-harlequin_r5d4_ready_blackbird.toddle.site/

### DIFF
--- a/packages/runtime/src/signal/signal.ts
+++ b/packages/runtime/src/signal/signal.ts
@@ -1,11 +1,8 @@
 import deepEqual from 'fast-deep-equal'
 
-const signals: Set<Signal<any>> = ((window as any).currentSignals =
-  (window as any).currentSignals ?? new Set<Signal<any>>())
-
 export class Signal<T> {
   value: T
-  subscribers: Array<{
+  subscribers: Set<{
     notify: (value: T) => void
     destroy?: () => void
   }>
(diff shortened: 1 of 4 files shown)


packages/runtime/src/signal/signal.ts
```diff
@@ -1,26 +1,28 @@
...
   set(value: T) {
+    // Short circuit and skip expensive `deepEqual` if there are not currently any subscribers
+    if (this.subscribers.size === 0) {
+      this.value = value
+      return
+    }
+
     if (deepEqual(value, this.value) === false) {
       this.value = value
       this.subscribers.forEach(({ notify }) => notify(this.value))
```
erik-beus, line 28: I guess this would be similar?
```suggestion
    this.value = value
    // Short circuit and skip expensive `deepEqual` if there are not currently any subscribers
    if (this.subscribers.size === 0) {
      return
    }

    if (deepEqual(value, this.value) === false) {
      this.subscribers.forEach(({ notify }) => notify(this.value))
```

erik-beus (commented):
Looks good, just 1 suggestion. I'll test 👍

PR #74, opened 2025-01-29
Allow relative destination url for routes

erik-beus:


mikkeltoddle (approved):

examples/hono/src/routes/routeHandler.ts
```diff
@@ -21,6 +27,9 @@ export const routeHandler = async (c: Context<HonoEnv>, route: Route) => {
     // unsupported accept headers from the client (brotli etc.)
     requestHeaders.set('accept-encoding', 'gzip')
     requestHeaders.set('accept', '*/*')
+    // Add header to identify that this is a rewrite
+    // This allows us to avoid recursive fetch calls across toddle routes
+    requestHeaders.set(REWRITE_HEADER, 'true')
```
mikkeltoddle, line 32: Should this be conditional? Or are redirects handled elsewhere?

mikkeltoddle (approved):

examples/hono/src/routes/routeHandler.ts
erik-beus: It now only happens for rewrites 👍

PR #75, opened 2025-01-29
Resolve contexts to any depth in preview 

jacob-kofoed:
Resolve contexts to any depth in preview and resolve variables after context values.

A previous fix was created where we calculate variables' initial value AFTER contexts are subscribed. This change fixes the same issue for the in-editor preview. However, it also fixes an issue where we only calculated context to one level instead of to any depth.

If component A use context from B, but some of B's exposed formulas rely on context from C, then C is first created (stub), then B and finally A can use closer to real context values.

Keep in mind all of these changes only matters when you are viewing a component in isolation, hence no real context is given. So all of the contexts are created from their test-values etc. We may at some point try running context providers' APIs,

### DIFF
--- a/packages/runtime/src/editor-preview.main.ts
+++ b/packages/runtime/src/editor-preview.main.ts
@@ -1053,6 +1053,88 @@ export const createRoot = (
         env,
       })
     }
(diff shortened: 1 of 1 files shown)


erik-beus (approved):
Looks good 🙌

PR #78, opened 2025-01-31
fix: add route parameters to component data

mikkeltoddle:
This PR will add Route parameters to component data, to support path and query parameters as input


packages/ssr/src/routing/routing.ts
```diff
@@ -99,10 +96,16 @@ export const getRouteDestination = ({
       // the route's source definition + global formulas.
       {
         data: {
-          'URL parameters': getDataUrlParameters({
-            route: route.source,
-            req,
-          }),
+          'Route parameters': {
+            path: route.source.path.reduce(
+              (prev, path) => ({ ...prev, [path.name]: path.testValue }),
```
erik-beus, line 101: We shouldn't be using the test values but the actual values coming from the request url :) Try looking at what `getDataUrlParameters` does perhaps

packages/core/src/component/component.types.ts
```diff
@@ -22,6 +22,10 @@ export interface ComponentData {
   Variables?: Record<string, unknown>
   Contexts?: Record<string, Record<string, unknown>>
   'URL parameters'?: Record<string, string | null>
+  'Route parameters': {
```
erik-beus, line 25: This must be optional since it's not available in all existing projects

erik-beus (changes requested):
Please see comments

packages/ssr/src/routing/routing.ts
mikkeltoddle: Yes! Makes sense.

packages/ssr/src/rendering/formulaContext.ts
```diff
@@ -136,29 +136,34 @@ const getParameters = ({
...
 
   // Explicitly set all query params to null by default
   // to avoid undefined values in the runtime
   const defaultQueryParams = Object.keys(route?.query ?? {}).reduce<
     Record<string, null>
   >((params, key) => ({ ...params, [key]: null }), {})
   return {
-    combinedParams: params,
-    hash: url.hash.slice(1),
+    pathParams: pathParams,
```
erik-beus, line 163:
```suggestion
    pathParams,
```

packages/ssr/src/routing/routing.ts
```diff
@@ -93,16 +93,24 @@ export const getRouteDestination = ({
...
       route.destination,
       // destination formulas should only have access to URL parameters from
       // the route's source definition + global formulas.
       {
         data: {
-          'URL parameters': getDataUrlParameters({
-            route: route.source,
-            req,
-          }),
+          'URL parameters': {
```
erik-beus, line 108: I guess this should follow your type change?
```suggestion
          'Route parameters': {
```

erik-beus (changes requested):
Please see comments

packages/ssr/src/routing/routing.ts
mikkeltoddle: 😬

packages/core/src/component/component.types.ts
```diff
@@ -22,6 +22,14 @@ export interface ComponentData {
   Variables?: Record<string, unknown>
   Contexts?: Record<string, Record<string, unknown>>
   'URL parameters'?: Record<string, string | null>
+  // eg. source: { path: { docs: null }, query: { embed: everything } }
+  'Route parameters'?: Record<
+    string,
```
erik-beus, line 27: What do we need the key for? I guess we will always only have 1 set of route parameters when evaluating a formula? If we only ever expect the key to be `source` I think we should be explicit about that in the type

erik-beus (changes requested):
See additional comment

packages/core/src/component/component.types.ts
```diff
@@ -22,6 +22,11 @@ export interface ComponentData {
   Variables?: Record<string, unknown>
   Contexts?: Record<string, Record<string, unknown>>
   'URL parameters'?: Record<string, string | null>
+  // eg. source: { path: { docs: null }, query: { embed: everything } }
```
erik-beus, line 25:
```suggestion
  // eg.: { path: { docs: null }, query: { embed: everything } }
```

erik-beus (approved):
:)

PR #80, opened 2025-02-03
Adjust rule to ensure unique path parameter names

erik-beus:
Our client-side routing would fail if two path parameters had the same name. See example here: 
- https://start-peach_roos_tarpals_wee_hoverfly.toddle.site/toddle/company
- https://toddle.dev/projects/peach_roos_tarpals_wee_hoverfly/branches/start/components/company?leftpanel=design&rightpanel=attributes&canvas-width=800&canvas-height=800

This change ensures that path parameters are unique while still ensuring that path and query parameters are named uniquely.

Closes https://github.com/toddledev/toddle-internal/issues/2191

### DIFF
--- a/package.json
+++ b/package.json
@@ -2,7 +2,7 @@
   "name": "toddle",
   "type": "module",
   "license": "Apache-2.0",
-  "version": "0.0.3-alpha.19",
+  "version": "0.0.3-alpha.20",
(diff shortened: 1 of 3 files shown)


jacob-kofoed (approved):
Makes sense

PR #81, opened 2025-02-03
Add support for serving custom elements

erik-beus:
To test, please run the hono example project and open the `examples/custom-element/custom-element.html` file in a browser.


examples/custom-element/custom-element.html
```diff
@@ -0,0 +1,17 @@
...
+  <head>
+    <meta charset="UTF-8" />
+    <title>Custom Element Test</title>
+  </head>
+  <body style="background-color: black;">
+    <h1 style="color: white;">Custom Element Test</h1>
+    <toddle-counter>
+      <script
+        type="module"
+        src="http://localhost:3030/.toddle/custom-element/counter.js"
```
jacob-kofoed, line 12: Maybe add a readme or comment to inform that this require hono to run with `template=medium` to work.

examples/custom-element/custom-element.html
```diff
@@ -0,0 +1,17 @@
...
+      ></script>
+    </toddle-counter>
```
jacob-kofoed, line 14: You can add any HTML inside a custom element by the way. You could add a `<strong>This text should be replaced by a counter component</strong>` etc.

examples/hono/src/routes/customElement.ts
```diff
@@ -0,0 +1,190 @@
...
+   .join('\n * ')}
+ *
+ * Events:
+ *
+ * ${Object.entries(component.events ?? {})
+   .map(([, event]) => `- ${event.name}`)
+   .join('\n * ')}
+ */
+
+import { defineComponents, loadCorePlugins } from '/_static/esm-custom-element.main.js';
```
jacob-kofoed, line 85: Only just noticed, but `esm-custom-element.main.js` is ~100kb. We really should check if we can reduce the size somehow. I is hard to suggest using custom-elements as a form of lazy-loading components if they have a 100+kb overhead.

packages/ssr/src/utils/tags.ts
```diff
@@ -0,0 +1,26 @@
+import type { Component } from '@toddledev/core/dist/component/component.types'
+
+/**
+ * Used to replace any recursive references to the root component, as it would be rendered both by toddle runtime
+ * and by the custom element itself (which would cause an infinite loop)
+ */
+export const replaceTagInNodes =
```
jacob-kofoed, line 7: Is this new or copy/paste from previously? I'm not sure I understand what it does (I'm aware that I may have written the code 😄)

erik-beus: You did write the code initially yes ;) It was just copied over from our internal repo 👍

examples/custom-element/custom-element.html
erik-beus: I only added the paragraph because the element wasn't rendering initially when I tested. I'll add an element inside of the custom element to show that it gets replaced 👍

PR #82, opened 2025-02-03
Fix an infinite loop in editor preview by keeping track of created context providers

jacob-kofoed:
I doubt anyone is actually using recursive context providers, but I just realised that it had a potential infinite loop. I tested it before/after with success.

### DIFF
--- a/packages/runtime/src/editor-preview.main.ts
+++ b/packages/runtime/src/editor-preview.main.ts
@@ -1054,12 +1054,19 @@ export const createRoot = (
       })
     }
     if (fastDeepEqual(_component.contexts, ctx?.component.contexts) === false) {
+      const contextProvidersCreated = new Set<string>()
       Contexts = (function createStaticContextFromComponent(
         component: Component,
       ) {
+        contextProvidersCreated.add(component.name)
         return mapObject(
           component.contexts ?? {},
           ([providerName, context]) => {
+            if (contextProvidersCreated.has(providerName)) {
+              // Circular dependency detected in context-providers (ie. A -> B -> A -> ...), stop recursion
+              return [providerName, {}]
+            }
+
             const providerComponent = getAllComponents().find(
               (c) => c.name === providerName,
             )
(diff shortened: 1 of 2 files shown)


packages/runtime/src/utils/BatchQueue.ts
```diff
@@ -3,22 +3,22 @@
  * This is more efficient than processing each callback in a separate requestAnimationFrame due to the overhead.
  */
 export class BatchQueue {
-  private batchQueue = new Set<() => void>()
+  private batchQueue: Array<() => void> = []
```
erik-beus, line 6: Your PR description doesn't mention why you changed this part? 😅

packages/runtime/src/editor-preview.main.ts
```diff
@@ -1054,12 +1054,19 @@ export const createRoot = (
       })
     }
     if (fastDeepEqual(_component.contexts, ctx?.component.contexts) === false) {
+      const contextProvidersCreated = new Set<string>()
       Contexts = (function createStaticContextFromComponent(
         component: Component,
       ) {
+        contextProvidersCreated.add(component.name)
```
erik-beus, line 1061: Should we perhaps include the package name as well if it's available? 🤔

erik-beus (changes requested):
Please see questions

packages/runtime/src/utils/BatchQueue.ts
jacob-kofoed: Ah sorry, I don't think it changes the logic in any way, but it is a few ms faster and makes more sense as it is based on native push/pop logic rather than anything Set provides extra.

packages/runtime/src/editor-preview.main.ts
jacob-kofoed: Our editor preview types are confusing me here. We don't have the package on the component, and the ctx is for the main component. However, the package name should already be prepended to the `component.name`: https://github.com/toddledev/toddle/blob/c8cb9dbf821584f0b148af7673287a2fc7b72c36/packages/runtime/src/editor-preview.main.ts#L436

I'm not sure why we ended with this solution, but it seems to work so far 😕

erik-beus (approved):

PR #91, opened 2025-02-06
Fix an issue where multiple context subscriptions using each other could lead to corrupted data

jacob-kofoed:
If `C -> A` and `B -> C` and `B -> A`, then A in B would not be mocked as A would already be created to populate the C mock. This fix an issue where contexts would not show up correctly in some more complex projects (editor only).

This still ensures that context subscriptions in a closed graph still don't infinitely create themselves (`A -> B -> A` or `A -> A`)

### DIFF
--- a/packages/runtime/src/editor-preview.main.ts
+++ b/packages/runtime/src/editor-preview.main.ts
@@ -1054,15 +1054,15 @@ export const createRoot = (
       })
     }
     if (fastDeepEqual(_component.contexts, ctx?.component.contexts) === false) {
-      const contextProvidersCreated = new Set<string>()
       Contexts = (function createStaticContextFromComponent(
         component: Component,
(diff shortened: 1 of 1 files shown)


packages/runtime/src/editor-preview.main.ts
```diff
@@ -1054,15 +1054,15 @@ export const createRoot = (
       })
     }
     if (fastDeepEqual(_component.contexts, ctx?.component.contexts) === false) {
-      const contextProvidersCreated = new Set<string>()
       Contexts = (function createStaticContextFromComponent(
         component: Component,
+        contextProvidersCreated?: Set<string>,
```
jacob-kofoed, line 1059: I parse this through the recursive function instead of having a single Set. We create a new base Set for each context formula.

erik-beus (approved):

PR #92, opened 2025-02-06
Improve Hono example with serveStatic middleware

erik-beus:
- Update README(s)
- Simplify Hono example
- Use dynamic import for loading the actual project


examples/hono/scripts/syncStaticAssets.js
```diff
@@ -0,0 +1,16 @@
+// Copy files from the static-assets directory to the dist directory using fs
+// This script is executed by the build process
+const fs = require('fs')
+const path = require('path')
+import { RESET_STYLES } from '@toddledev/core/dist/styling/theme.const'
+;[
+  'esm-page.main.js',
```
jacob-kofoed, line 7: FYI I think the usual naming convention is to add the "esm" just before the file type: `page.main.esm.js` etc.

jacob-kofoed (approved):

examples/hono/scripts/syncStaticAssets.js
erik-beus: Good point. I'll adjust that in a future PR 👍

PR #97, opened 2025-02-07
Fix/no reference event rule

marija04:
This will fix the error when seeing an issue about event not used in a component when actually it's used but triggered by the legacy event trigger.


packages/search/src/util/shouldSearchPath.ts
```diff
@@ -9,3 +11,19 @@ export function shouldSearchPath(
     )
   )
 }
+
+export const isLegacyAction = (model: ActionModel) => {
```
jacob-kofoed, line 15: Not sure this belongs in this file. Maybe move to `utils/isLegacyAction.ts`

packages/search/src/util/shouldSearchPath.ts
```diff
@@ -9,3 +11,19 @@ export function shouldSearchPath(
...
+      return legacyCustomActions.some(
+        (action) => action.name === model.name && action.type === model?.type,
+      )
+  }
+  return false
+}
+
+const legacyCustomActions = [
+  { type: undefined, name: 'If' },
+  { type: undefined, name: 'TriggerEvent' },
```
jacob-kofoed, line 28: We have many more legacy actions by the way: `StopPropagation`, `PreventDefault` etc.

packages/search/src/rules/noReferenceEventRule.test.ts
```diff
@@ -107,6 +134,7 @@ describe('noReferenceEventRule', () => {
       }),
     )
 
+    console.log('problems', problems)
```
jacob-kofoed, line 137: remove log before merging

packages/search/src/util/shouldSearchPath.ts
marija04: I actually didn't notice the name of the file 🙈 
Wouldn't be better to have all this helpful functions in one file?

packages/search/src/util/shouldSearchPath.ts
marija04: I moved this function from another file and just added the `TriggerEvent`. Do we need to add all of the other legacy actions?

packages/search/src/rules/legacyActionRule.ts
marija04, line 1: @jacob-kofoed do you think this rule will still be correct now that I have added `TriggerEvent` to the `legacyCustomActions`?

bun.lock
```diff
@@ -1,5 +1,5 @@
 {
-  "lockfileVersion": 1,
+  "lockfileVersion": 0,
```
erik-beus, line 2: You should upgrade your bun version to the latest version

erik-beus (changes requested):

packages/search/src/util/shouldSearchPath.ts
jacob-kofoed: Sorry for late response, I sometimes drown a bit in github notifications (feel free to bump me). I just added the other legacy actions in this PR https://github.com/toddledev/toddle/pull/105 so you may have some conflicts here.

packages/search/src/rules/noReferenceEventRule.ts
```diff
@@ -13,11 +14,26 @@ export const noReferenceEventRule: Rule<{ name: string }> = {
     const events = memo(`${component.name}-events`, () => {
       const events = new Set<string>()
       for (const [, action] of component.actionModelsInComponent()) {
-        if (action.type === 'TriggerEvent') {
+        if (isLegacyAction(action)) {
+          if (
+            'name' in action &&
+            'arguments' in action &&
+            action.name === 'TriggerEvent'
+          ) {
```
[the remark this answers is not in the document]
jacob-kofoed: Just checked. The TriggerEvent will not have a version field, so you can check that it is undefined. Any custom action created by users should have a `version: 2+`

jacob-kofoed (approved):
Looks good to me! If you did a quick test to see number of issues is correct, then you can just go ahead and merge.

erik-beus (approved):

PR #105, opened 2025-02-12
Populate legacy action rule

jacob-kofoed:
Very toddle project-specific rule, but hopefully, it is just a temporary rule until we fix the ~900 legacy actions.

### DIFF
--- a/packages/search/src/rules/legacyActionRule.ts
+++ b/packages/search/src/rules/legacyActionRule.ts
@@ -15,8 +15,11 @@ export const legacyActionRule: Rule<{
     if (isLegacyAction(value)) {
       let details: { name: string } | undefined
       if ('name' in value) {
-        details = { name: value.name }
+        details = {
+          name: value.name,
+        }
       }
+
       report(path, details)
     }
   },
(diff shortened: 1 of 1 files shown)


packages/search/src/rules/legacyActionRule.ts
```diff
@@ -26,11 +29,28 @@ const isLegacyAction = (model: ActionModel) => {
...
+  'Update URL parameter',
+  'updateUrlParameters',
+  'UpdateQueryParam',
+  'Update Query',
+  'Fetch',
+  'SetTimeout',
+  'SetInterval',
+  'FocusElement',
+  'Debug',
+  'GoToURL',
```
erik-beus, line 55: I believe the `TriggerEvent` action should also be considered a legacy action?
```suggestion
  'GoToURL',
  'TriggerEvent',
```

erik-beus (changes requested):
Please see suggestion

packages/search/src/rules/legacyActionRule.ts
jacob-kofoed: Very true! I committed your suggestion. I tried to be thorough, but might have missed something still.

PR #106, opened 2025-02-12
Fix an issue where setting an URL param would clear all others

jacob-kofoed:
This is an editor-only inconsistency, it already works as expected in preview mode.

Fixes https://github.com/toddledev/toddle-internal/issues/2168

I changed it to use the location signal directly to not handle state twice in UrlParameters and locationSignal. This is important as we need to set the initial values to their test values (before we used a fallback method), but when the location signal updates we need to keep previously set values. There was no way to do this before and we just set them all (except the changed ones) to undefined. The locationSignal would be empty initially and only contain changed values.

### DIFF
--- a/packages/runtime/src/editor-preview.main.ts
+++ b/packages/runtime/src/editor-preview.main.ts
@@ -1179,7 +1179,6 @@ export const createRoot = (
         'URL parameters': {
           ...window.toddle.locationSignal.get().query,
           ...window.toddle.locationSignal.get().params,
-
         } as Record<string, string>,
         Attributes,
         Variables,


packages/runtime/src/editor-preview.main.ts
```diff
@@ -1167,7 +1176,10 @@ export const createRoot = (
     dataSignal.update((data) => {
       return {
         ...data,
-        'URL parameters': UrlParameters,
+        'URL parameters': {
+          ...window.toddle.locationSignal.get().params,
+          ...window.toddle.locationSignal.get().query,
```
erik-beus, line 1181: I believe we prefer path parameters over query parameters in our routing logic in case of naming conflict?
```suggestion
          ...window.toddle.locationSignal.get().query,
          ...window.toddle.locationSignal.get().params,

```

erik-beus (commented):
Please see comment

erik-beus (approved):

PR #110, opened 2025-02-14
Support SVG text tag

jacob-kofoed:
This also fixes an issue where only the first text-node inside script and style tags would be rendered.

Handling text differently for some elements is a temporary measure until we merge (we have a POC) a change to skip the span around text-nodes, which should also greatly reduce the DOM element count.

Can be tested here https://svgtext-jacob_test.toddle.site/ uses SVG and should render a style tag combined of three text nodes where one is dynamic to change the background.

### DIFF
--- a/packages/runtime/src/components/createElement.ts
+++ b/packages/runtime/src/components/createElement.ts
@@ -1,4 +1,4 @@
-import {
+import type {
   ElementNodeModel,
   NodeModel,
 } from '@toddledev/core/dist/component/component.types'
@@ -13,7 +13,7 @@ import type { Signal } from '../signal/signal'
 import { getDragData } from '../utils/getDragData'
 import { getElementTagName } from '../utils/getElementTagName'
 import { setAttribute } from '../utils/setAttribute'
-import { NodeRenderer, createNode } from './createNode'
+import type { NodeRenderer, createNode } from './createNode'
 
 export function createElement({
   node,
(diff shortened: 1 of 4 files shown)


packages/runtime/src/components/createElement.ts
```diff
@@ -1,4 +1,7 @@
-import { ElementNodeModel } from '@toddledev/core/dist/component/component.types'
+import {
```
erik-beus, line 1:
```suggestion
import type {
```

packages/runtime/src/components/createElement.ts
```diff
@@ -158,35 +161,52 @@ export function createElement({
     elem.addEventListener(event.trigger, handler)
   })
 
-  // for script and style tags we just render the first text child.
+  // for script, style & SVG<text> tags we only render text child.
```
erik-beus, line 164: It probably doesn't matter, but you could avoid (potentially) calling the same function 3 x by extracting it to a const. Also, I suggest updating the comment
```suggestion
  // for script, style & SVG<text> tags we only render text children
  const nodeTag = node.tag.toLocaleLowerCase()
```

erik-beus (changes requested):
Please see suggestions

packages/runtime/src/components/createElement.ts
jacob-kofoed: Did you research any way to enforce this during typecheck? Otherwise I'll give it a go.

packages/runtime/src/components/createElement.ts
jacob-kofoed: good suggestion. I wanted to do a string array, but `text` is dependent on it being an SVG, so it is more logic than a static lookup. Your suggestion helps on readablity 👍

packages/runtime/src/components/createElement.ts
jacob-kofoed: I added `https://typescript-eslint.io/rules/consistent-type-imports/` to avoid these kinds of issues in the future. I did it as a warning for now as I am too lazy to fix all the potential issues right now, but I will look into a follow-up PR and turn it to an error.

erik-beus (approved):

PR #115, opened 2025-02-18
Add staticSizeConstraint rule

erik-beus:
WIP


packages/search/src/rules/issues/miscellaneous/miscRules.index.ts
```diff
@@ -1,5 +1,11 @@
...
 import { noReferenceNodeRule } from './noReferenceNodeRule'
 import { requireExtensionRule } from './requireExtensionRule'
 import { unknownCookieRule } from './unknownCookieRule'
 
-export default [noReferenceNodeRule, requireExtensionRule, unknownCookieRule]
+export default [
+  noReferenceNodeRule,
+  requireExtensionRule,
+  unknownCookieRule,
+  createStaticSizeConstraintRule('svg', 100 * 1024), // 100 KB is a large SVG
```
jacob-kofoed: Should we add `img` right away, to catch base64 encoded massive images?

erik-beus: Good idea 👍

erik-beus: Updated

packages/search/src/rules/issues/miscellaneous/createStaticSizeConstraintRule.ts
```diff
@@ -0,0 +1,76 @@
...
+          }
+        }
+        const staticElement = evaluateElement(args.value)
+        size = new Blob([staticElement]).size
+        if (size > maxSize) {
+          const formatNumber = (num: number) => Intl.NumberFormat().format(num)
+          report({
+            path: args.path,
+            info: {
+              title: `Element size exceeds the suggested maximum allowed size of ${formatNumber(maxSize)} bytes`,
```
jacob-kofoed: Maybe add (vibe-code) a byte formatter, so we get 100 kilobytes instead of 100.000 bytes here?

erik-beus: Good idea. Output now looks like this:
```
{
  title: "Element size exceeds the suggested maximum allowed size of 1 byte",
  description: "The <img> element has a size of 1.17 KB, which exceeds the suggested limit of 1 byte. Consider simplifying the content or structure of this element to reduce its size and improve performance.",
}
```

PR #116, opened 2025-02-18
Support nullable path in route declaration

erik-beus:
When a new page is created in the editor, it doesn'a have any path parameters if no path is defined. That currently leads to an error 500

Here's an example project https://start-fuchsia_saesee_tiin_associated_junglefowl.toddle.site/

See reported issue here https://discord.com/channels/972416966683926538/1341416733772353660

### DIFF
--- a/packages/ssr/src/routing/routing.ts
+++ b/packages/ssr/src/routing/routing.ts
@@ -1,7 +1,7 @@
 import { getUrl } from '@toddledev/core/dist/api/api'
 import {
   PageComponent,
-  PageRoute,
+  RouteDeclaration,
 } from '@toddledev/core/dist/component/component.types'
 import { isDefined } from '@toddledev/core/dist/utils/util'
 import {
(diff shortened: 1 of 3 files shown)


packages/ssr/src/routing/routing.ts
```diff
@@ -43,15 +43,16 @@ export const matchRoutes = <T>({
...
 }): T | undefined => {
+  console.log('entries', entries)
   const pathSegments = getPathSegments(url)
   const matches = Object.values(entries)
     .filter((entry) => {
       const route = getRoute(entry)
       return (
-        pathSegments.length <= route.path.length &&
-        route.path.every(
+        pathSegments.length <= (route.path ?? []).length &&
```
mikkeltoddle, line 54: How about we replace the getRoute input (which is always the same method) with a reusable getRoute function that returns a defined path? Then you could also use that in formulaPath?

erik-beus:
We'll fix in the editor instead 👍

PR #126, opened 2025-02-21
Pass namespace through components and support MathML node namespace

jacob-kofoed:
This change allows us to split a single SVG (or MathML) into multiple components. Previously the current namespace was lost when moving into child components.

In this PR, I also extend toddle to support the MathML namespace. There does exist more namespaces, but MathML is already a pretty obscure use case.

When versioning is supported, I suggest we explicitly set the namespace if a node has an `xmlns` attribute instead of only inferring it from the tag name.

### DIFF
--- a/packages/runtime/src/components/createElement.ts
+++ b/packages/runtime/src/components/createElement.ts
@@ -22,14 +22,24 @@ export function createElement({
   id,
   path,
   ctx,
-  isSvg,
+  namespace,
   instance,
 }: NodeRenderer<ElementNodeModel>): Element {
   const tag = getElementTagName(node, ctx, id)
-  const elem =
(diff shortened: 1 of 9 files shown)


packages/runtime/src/components/createElement.ts
```diff
@@ -218,7 +227,7 @@ export function createElement({
         dataSignal,
         ctx,
         // eslint-disable-next-line @typescript-eslint/prefer-nullish-coalescing
-        isSvg: isSvg || tag === 'svg',
+        namespace,
```
erik-beus, line 230: You can now remove line 229

erik-beus (approved):
Looks good. Just 1 small note

PR #133, opened 2025-02-25
fix: add open in new tab support to go to url

mikkeltoddle:
This will add support to navigate in a new tab.

### DIFF
--- a/packages/lib/actions/gotToURL/action.json
+++ b/packages/lib/actions/gotToURL/action.json
@@ -8,6 +8,13 @@
         "type": "value",
         "value": "https://example.com"
       }
+    },
+    {
+      "name": "Open in new tab",
+      "formula": {
+        "type": "value",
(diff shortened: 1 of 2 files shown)


packages/lib/actions/gotToURL/action.json
```diff
@@ -8,6 +8,13 @@
         "type": "value",
         "value": "https://example.com"
       }
+    },
+    {
+      "name": "Open in new tab",
```
erik-beus, line 13: It would be more flexible if we called this parameter `target` and allowed passing one of the valid target properties? See https://developer.mozilla.org/en-US/docs/Web/API/Window/open#target

While it does mean that users will need to know what a valid `target` is, it also helps them learn the web platform ;)

erik-beus (changes requested):
Please see suggestion

PR #134, opened 2025-02-25
Add unknownWorkflowParameterRule

erik-beus:
- Detect unknown workflow parameter rules in component workflows
- Detect unknown workflow parameter rules in context workflows
- Detect unknown workflow parameter rules in context workflows from packages
- Add tests
- Restructure workflow tests into their own folder

### DIFF
--- a/packages/search/src/rules/workflows/unknownWorkflowParameterRule.ts
+++ b/packages/search/src/rules/workflows/unknownWorkflowParameterRule.ts
@@ -1,0 +1,55 @@
+import type { ComponentWorkflow } from '@toddledev/core/dist/component/component.types'
+import type { Rule } from '../../types'
+
+export const unknownWorkflowParameterRule: Rule<{ parameter: string }> = {
+  code: 'unknown workflow parameter',
+  level: 'error',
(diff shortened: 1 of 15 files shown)


packages/search/src/rules/workflows/unknownWorkflowParameterRule.ts
```diff
@@ -0,0 +1,55 @@
...
+      }
+      workflow = subscription?.package
+        ? files.packages?.[subscription.package]?.components[
+            subscription.componentName
+          ]?.workflows?.[value.workflow]
+        : files.components[subscription.componentName]?.workflows?.[
+            value.workflow
+          ]
+      if (!workflow) {
+        console.error('Unable to find workflow')
```
jacob-kofoed, line 35: Not sure if we should log as error considering we have other checks for unknown workflow subscriptions. Lets keep it, but if it spams the console while also showing an issue-rule, then maybe we should just fail quietly.

jacob-kofoed (approved):
Looks good; remember to add the error message to the toddle project before going live

jacob-kofoed (approved):

PR #136, opened 2025-02-25
Fix all type imports

erik-beus:
Prefer type imports since they're more explicit and probably easier to tree-shake


jacob-kofoed (approved):
Looks correct. How did you auto-fix all these issues?

I doubt it, but we should check if this had any impact on bundle size at all.

erik-beus:
> Looks correct. How did you auto-fix all these issues?
> 
> I doubt it, but we should check if this had any impact on bundle size at all.

I tried to run `bunx eslint --fix --rule '@typescript-eslint/consistent-type-imports:[error]'` - but apparently, the `--fix` parameter doesn't work for a single rule 🤷‍♂️ So I went through all files and auto fixed where I could using the eslint browser plugin 😅

PR #146, opened 2025-02-26
Fix/api nullable inputs

ChrEsb:


packages/core/src/api/apiTypes.ts
```diff
@@ -62,7 +62,7 @@ export interface ApiRequest extends ApiBase {
   method?: ApiMethod
   body?: Formula
   // inputs for an API request - the server will only accept listed arguments
```
erik-beus, line 64:
```suggestion
  // inputs for an API request
```

packages/runtime/src/api/createAPIv2.ts
```diff
@@ -80,7 +80,7 @@ export function createAPI(
     const evaluatedInputs = Object.entries(api.inputs).reduce<
       Record<string, unknown>
     >((acc, [key, value]) => {
-      acc[key] = applyFormula(value.formula, formulaContext)
+      acc[key] = applyFormula(value.formula ?? undefined, formulaContext)
```
erik-beus, line 83: We usually try to avoid `undefined` in the editor. Would it make sense to use `null` instead here?
```suggestion
      acc[key] = applyFormula(value.formula ?? null, formulaContext)
```

packages/runtime/src/events/handleAction.ts
```diff
@@ -124,7 +124,7 @@ export function handleAction(
         // This is needed if the inputs are formulas referencing workflow parameters
         const actionInputs = isv2
           ? mapValues(action.inputs ?? {}, (input) =>
-              applyFormula(input.formula, {
+              applyFormula(input.formula ?? undefined, {
```
erik-beus, line 127:
```suggestion
              applyFormula(input.formula ?? null, {
```

erik-beus (changes requested):
Please see suggestions

erik-beus (approved):

PR #155, opened 2025-03-04
Load external page stylesheet from new endpoint

erik-beus:
Currently, we inline CSS for all components in the page. To improve the browser's chance of caching the CSS, reduce the page (HTML document) size and improve the performance of the page, we should load the CSS from a separate endpoint.

This PR:
- Adds a link element in the `<head>` that loads an external stylesheet
- Adds an endpoint in the hono example that serves the stylesheet (uncached in the example)
- Stops inlining the component styles
- Reduces the amount of whitespace in the style output

### DIFF
--- a/examples/cloudflare-worker/src/index.ts
+++ b/examples/cloudflare-worker/src/index.ts
@@ -11,6 +11,7 @@ import { manifest } from './routes/manifest'
 import { robots } from './routes/robots'
 import { serviceWorker } from './routes/serviceWorker'
 import { sitemap } from './routes/sitemap'
+import { stylesheetHandler } from './routes/stylesheetHandler'
 import { toddlePage } from './routes/toddlePage'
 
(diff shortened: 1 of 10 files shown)


jacob-kofoed:
This fits perfectly into the idea of splitting the final output into smaller pieces for better caching and reuse across pages. This is a must if we want MPA to be a thing, any duplicate data should be split into their own file.

With enough testing we can merge this without versioning imo.

erik-beus:
@jacob-kofoed @ChrEsb this change would mean a page would need to load 2 stylesheets before it looks correct:
- The reset stylesheet (ideally first)
- The page specific stylesheet

Alternatively, we combine the 2 and serve them as 1 stylesheet. That would mean that we can't cache the reset stylesheet across pages though.
What are your thoughts?

ChrEsb (approved):

PR #160, opened 2025-03-05
Do not report global formula/actions that are exported from packages

jacob-kofoed:
We already had a similar check for component, but were missing from formulas/actions.


marija04 (approved):
Looks good 🙂

PR #177, opened 2025-03-12
Added logarithm formula

Xeiik:
Added a simple formula that returns the natural logarithm of a number ( `Math.log()` ).

### DIFF
--- a/packages/lib/formulas/logarithm/handler.test.ts
+++ b/packages/lib/formulas/logarithm/handler.test.ts
@@ -1,0 +1,11 @@
+import { describe, expect, test } from '@jest/globals'
+import handler from './handler'
+
+describe('Formula: Logarithm', () => {
+  test('should return the logarithm of the number', () => {
+    expect(handler([-1], undefined as any)).toBe(NaN)
+    expect(handler([0], undefined as any)).toBe(-Infinity)
+    expect(handler([1], undefined as any)).toBe(0)
+    expect(handler([10], undefined as any)).toBe(2.302585092994046)
+  })
+})
--- a/packages/lib/formulas/logarithm/handler.test.ts
+++ b/packages/lib/formulas/logarithm/handler.test.ts
@@ -3,7 +3,7 @@ import handler from './handler'
 
 describe('Formula: Logarithm', () => {
   test('should return the logarithm of the number', () => {
-    expect(handler([-1], undefined as any)).toBe(NaN)
+    expect(handler([-1], undefined as any)).toBe(null)
     expect(handler([0], undefined as any)).toBe(-Infinity)
     expect(handler([1], undefined as any)).toBe(0)
     expect(handler([10], undefined as any)).toBe(2.302585092994046)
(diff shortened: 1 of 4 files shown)


packages/lib/formulas/logarithm/handler.ts
```diff
@@ -0,0 +1,4 @@
+import type { FormulaHandler } from '@toddledev/core/dist/types'
+
+const handler: FormulaHandler<number> = ([a]) => Math.log(a)
```
erik-beus, line 3: I suggest it always returns a number or `null` since we usually don't work with `NaN` in toddle, so:
- `log(-1) -> null`
- `log(null) -> null`
- `log("hello") -> null`

That would also resemble the pattern we use in other formulas like this one https://github.com/toddledev/toddle/blob/main/packages/lib/formulas/add/handler.ts

erik-beus (changes requested):
Thank you so much for your contribution @Xeiik 🙌 

Please see my suggestion.
Also, the tests seem to succeed, but running the `typecheck` command in the `lib/` package fails atm.

packages/lib/formulas/logarithm/handler.ts
Xeiik: Makes sense, I updated the function to check for type and negative values.

packages/lib/formulas/logarithm/handler.ts
```diff
@@ -0,0 +1,16 @@
+import type { FormulaHandler } from '@toddledev/core/dist/types'
+
+const handler: FormulaHandler<number> = ([a]) => {
+  // Check if input is not a number or is null
+  if (typeof a !== 'number' || a === null || Number.isNaN(a)) {
```
erik-beus, line 5: Technically, the second condition is obsolete, since `a` can't be a `null` if its type is `number` ;)
```suggestion
  if (typeof a !== 'number' || Number.isNaN(a)) {
```

packages/lib/formulas/logarithm/handler.test.ts
```diff
@@ -0,0 +1,11 @@
+import { describe, expect, test } from '@jest/globals'
+import handler from './handler'
+
+describe('Formula: Logarithm', () => {
+  test('should return the logarithm of the number', () => {
+    expect(handler([-1], undefined as any)).toBe(null)
+    expect(handler([0], undefined as any)).toBe(-Infinity)
+    expect(handler([1], undefined as any)).toBe(0)
+    expect(handler([10], undefined as any)).toBe(2.302585092994046)
```
erik-beus, line 9: It would be great if you could a few more test cases for `null`, `undefined` and a `string` perhaps:
```suggestion
    expect(handler([10], undefined as any)).toBe(2.302585092994046)
    expect(handler([null], undefined as any)).toBe(null)
    expect(handler([], undefined as any)).toBe(null)
    expect(handler(["hello"], undefined as any)).toBe(null)
```

erik-beus (changes requested):
Please see suggestions

packages/lib/formulas/logarithm/handler.test.ts
Xeiik: One day I'll be as good as the god of code

erik-beus (approved):

PR #178, opened 2025-03-12
Allow div as a direct child of dl

erik-beus:
See https://developer.mozilla.org/en-US/docs/Web/HTML/Element/dl#wrapping_name-value_groups_in_div_elements

Also, see reported issue here https://discord.com/channels/972416966683926538/1348724790013464596


jacob-kofoed (approved):
The more you know

PR #183, opened 2025-03-17
Attach assets when creating a release

erik-beus:


bin/prepareReleaseAssets.ts
```diff
@@ -0,0 +1,61 @@
...
+  ])
+
+  await bundleFiles(
+    [
+      'packages/runtime/src/page.main.ts',
+      'packages/runtime/src/editor-preview.main.ts',
+      'packages/runtime/src/custom-element.main.ts',
+      'packages/core/src/component/ToddleComponent.ts',
+      'packages/core/src/formula/ToddleFormula.ts',
+      'packages/core/src/api/api.ts',
```
jacob-kofoed, line 51: What is `api.ts` used for?

erik-beus: We use it in the editor for getting the url for an API request. This was to avoid replicating the logic in a custom formula.

PR #188, opened 2025-03-20
Support historyMode in SetURLParameterAction

erik-beus:
To allow more flexibility when updating path/query parameters, we want to allow setting whether the action should push or replace the history state. This is also useful when you want to update multiple parameters - although we should have a separate/similar action for that scenario.

### DIFF
--- a/packages/runtime/src/events/handleAction.ts
+++ b/packages/runtime/src/events/handleAction.ts
@@ -84,18 +84,15 @@ export function handleAction(
       }
       case 'SetURLParameter': {
         ctx.toddle.locationSignal.update((current) => {
-          // We only evaluate the formula if the parameter is
-          // available in the page's route
-          const getValue = () =>
-            applyFormula(action.data, {
-              data,
-              component: ctx.component,
-              formulaCache: ctx.formulaCache,
-              root: ctx.root,
-              package: ctx.package,
-              toddle: ctx.toddle,
-              env: ctx.env,
-            })
+          const value = applyFormula(action.data, {
+            data,
+            component: ctx.component,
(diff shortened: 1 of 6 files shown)


packages/runtime/src/page.main.ts
```diff
@@ -134,44 +134,6 @@ export const createRoot = (domNode: HTMLElement) => {
...
-        path = '/' + pathSegments.join('/')
-      } else {
-        path = compile(page as string, { encode: encodeURIComponent })(params)
-      }
-      const hashString = hash === undefined || hash === '' ? '' : '#' + hash
-      const queryString = Object.entries(query)
-        .filter(([_, q]) => q !== null)
-        .map(([key, value]) => {
-          return `${encodeURIComponent(
-            component?.route?.query[key]?.name ?? key,
```
erik-beus, line 164: Note that it uses `component` here (which is declared in `pate.main.ts`) - not `route` from the signal. I've changed that during the refactor

packages/runtime/src/events/handleAction.ts
```diff
@@ -91,23 +93,54 @@ export function handleAction(
...
             }
           }
+          if (!historyMode) {
+            // No path/query parameter matched
+            return current
+          }
+
+          const currentUrl = getLocationUrl(current)
+          const historyUrl = getLocationUrl(newLocation)
+          if (historyUrl !== currentUrl) {
```
erik-beus, line 132: This is the same behaviour as previously, since the `urlSignal.subscribe` function would only trigger when the url changed

jacob-kofoed (approved):
Nice changes, no comments. Works for me ✅

PR #191, opened 2025-03-21
Fix/custom event falsy value

ChrEsb:
Fixed an issue where the event data wasn't passed on correctly to a callback from a custom event if the value of the data was falsy.


packages/runtime/src/events/handleAction.ts
```diff
@@ -244,9 +244,11 @@ export function handleAction(
               subEvent.actions.forEach((action) =>
                 handleAction(
                   action,
-                  eventData
```
erik-beus, line 247: Could we perhaps introduce/modify a linter rule to catch this going forward? It would be safer if we could only compare on booleans?

erik-beus (approved):

PR #193, opened 2025-03-21
Add legacyApiRule and nonEmptyVoidElementRule

erik-beus:
- Restructure most rules into relevant folders
- Add `legacyApiRule` to encourage the new APIs
- Add `nonEmptyVoidElementRule` to highlight void elements that have children declared


packages/search/src/rules/dom/nonEmptyVoidElementRule.ts
```diff
@@ -0,0 +1,23 @@
+import type { Rule } from '../../types'
+
+import { VOID_HTML_ELEMENTS } from '@toddledev/ssr/dist/const'
+/**
+ * Lighthouse reports a similar issue:
+ * https://web.dev/articles/optimize-cls?utm_source=lighthouse&utm_medium=devtools#images_without_dimensions
```
jacob-kofoed, line 6: Nice, until we have our own docs, perhaps you can link to this web.dev page from the issues rule.

erik-beus: The link was from a copy/paste mistake 🙈 I've corrected it, but will add the new link in the issue description in the editor as well 👍

PR #202, opened 2025-03-26
Structure and editor messages for handling animations

jacob-kofoed:
Animations will be stored on nodes for now. I hope to have it stored on a styles object v2 one day.

The editor can now communicate some new messages, like emitting the computed styles for a node and setting a keyframe animation dynamically.

This requires a commit to toddle-internal to set up some styling in the head (the editor global styling should probably live in the toddle repo 🤔)

### DIFF
--- a/packages/core/src/styling/variantSelector.ts
+++ b/packages/core/src/styling/variantSelector.ts
@@ -1,5 +1,8 @@
 import type { CSSProperties } from 'react'
-import type { EventModel } from '../component/component.types'
+import type {
+  AnimationKeyframe,
+  EventModel,
+} from '../component/component.types'
 import type { Formula } from '../formula/formula'
 
 export type Shadow = {
@@ -98,6 +101,7 @@ export type ElementNodeModel = {
   attrs: Record<string, Formula>
   style: NodeStyleModel
   variants?: StyleVariant[]
+  animations?: Record<string, Record<string, AnimationKeyframe>>
   children: NodeModel[]
   events: EventModel[]
 }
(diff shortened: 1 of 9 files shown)


packages/core/src/styling/variantSelector.ts
```diff
@@ -98,6 +101,7 @@ export type ElementNodeModel = {
   attrs: Record<string, Formula>
   style: NodeStyleModel
   variants?: StyleVariant[]
+  animations?: Record<string, Record<string, AnimationKeyframe>>
```
erik-beus, line 104: The `ElementNodeModel` and `ComponentNodeModel` types in this file seem almost duplicate of the ones in `packages/core/src/component/component.types.ts`? 🤔

packages/runtime/src/styles/style.ts
```diff
@@ -108,6 +108,29 @@ ${
         .join('\n')
     : ''
 }
+
+${
+  node.animations
```
erik-beus, line 113: This seems similar to the addition in `packages/core/src/styling/style.css.ts`?

erik-beus (changes requested):
Please see questions. Would be nice reduce code duplication if possible 🤞

packages/runtime/src/styles/style.ts
jacob-kofoed: One is for the canvas and the other for live sites, as I understand it 🤔 Am I misunderstanding something perhaps? We should unite them, but I assume there is a reason that we have both at the moment, not sure what it is thought.

erik-beus (approved):

PR #206, opened 2025-03-29
Remove more test data from components

erik-beus:
We were not removing:

- Test data from workflow parameters
- Test data from formula inputs
- Service references from APIs
  
I also improved the type safety of the `omitKeys` utility function. I tried to adjust the `omit` function as well, but since we sometimes use it recursively, it's difficult to keep type safety.

### DIFF
--- a/packages/ssr/src/rendering/testData.ts
+++ b/packages/ssr/src/rendering/testData.ts
@@ -1,5 +1,10 @@
+import { isLegacyApi } from '@toddledev/core/dist/api/api'
 import type { Component } from '@toddledev/core/dist/component/component.types'
-import { mapObject, omit } from '@toddledev/core/dist/utils/collections'
+import {
+  mapObject,
+  omit,
+  omitKeys,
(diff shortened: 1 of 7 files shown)


packages/ssr/src/rendering/testData.ts
```diff
@@ -20,5 +25,38 @@ export function removeTestData(component: Component): Component {
...
+                omit(p, ['testValue']),
+              ),
+            },
+          ]),
+        }
+      : {}),
+    // eslint-disable-next-line @typescript-eslint/no-unnecessary-condition
+    apis: mapObject(component.apis ?? {}, ([key, api]) => [
+      key,
+      isLegacyApi(api) ? api : omitKeys(api, ['service', 'servicePath']),
```
jacob-kofoed, line 59: Just for knowledge sharing. I suppose service and servicePath are just for editor use? The actual URL etc. is all in the API.

jacob-kofoed (approved):

PR #208, opened 2025-03-31
Implemented preview_style canvas message type

jacob-kofoed:
- Added preview_style message type
- Actually clean up styles when switching components in editor
- Improve editor performance when setting styles

### DIFF
--- a/packages/runtime/src/editor-preview.main.ts
+++ b/packages/runtime/src/editor-preview.main.ts
@@ -6,6 +6,8 @@ import type {
   AnimationKeyframe,
   Component,
   ComponentData,
+  ComponentNodeModel,
+  ElementNodeModel,
   MetaEntry,
 } from '@toddledev/core/dist/component/component.types'
 import type {
@@ -15,6 +17,7 @@ import type {
 import { applyFormula } from '@toddledev/core/dist/formula/formula'
 import type { PluginFormula } from '@toddledev/core/dist/formula/formulaTypes'
 import { valueFormula } from '@toddledev/core/dist/formula/formulaUtils'
+import { getClassName } from '@toddledev/core/dist/styling/className'
 import type { OldTheme, Theme } from '@toddledev/core/dist/styling/theme'
 import { getThemeCss } from '@toddledev/core/dist/styling/theme'
 import { theme } from '@toddledev/core/dist/styling/theme.const'
@@ -121,6 +124,7 @@ type ToddlePreviewEvent =
         | undefined
       fillMode: 'none' | 'forwards' | 'backwards' | 'both' | undefined
     }
+  | { type: 'preview_style'; styles: Record<string, string> | null }
 
 /**
  * Styles required for rendering the same exact text again somewhere else (on a overlay rect in the editor)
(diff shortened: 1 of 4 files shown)


packages/runtime/src/editor-preview.main.ts
```diff
@@ -368,6 +373,12 @@ export const createRoot = (
             return
           }
           if (message.data.component.name != component?.name) {
+            // Clean up all styles before switching components
+            document.head
```
jacob-kofoed, line 377: We never actually cleaned styles when switching between components in the editor. They just kept adding up. Technically it could improve performance slightly when switching back to the same component again, but I don't think the overhead of leaking so many nodes and styles adds up to it.

packages/runtime/src/editor-preview.main.ts
```diff
@@ -1372,39 +1421,93 @@ export const createRoot = (
...
-        ctx: newCtx,
-        parentElement: domNode,
-        instance: { [newCtx.component.name]: 'root' },
+      // Remove preview styles
+      document.head.querySelector('[data-id="selected-node-styles"]')?.remove()
+
+      // Is only style change, no need to re-render
+      // Remove style and styleVariants from each node
+      const newComponentWithoutStyles = structuredClone(newCtx.component)
+      Object.values(newComponentWithoutStyles.nodes).forEach((node) => {
```
jacob-kofoed, line 1431: We would run an expensive update whenever ANYTHING changed in the active component. This change makes it so we skip the update and only run updateStyles when only styles change.

packages/runtime/src/styles/style.ts
```diff
@@ -160,15 +157,7 @@ ${
...
-          insertComponentStyles(childComponent, node.package ?? package_name) // write the dependant CSS first.
-          parent.querySelector(`[data-hash="${id}"]`)?.remove() //  remove the old styles
-          parent.appendChild(
-            getNodeStyles(
-              node,
-              toValidClassName(`${component.name}:${id}`, true),
-            ),
-          )
-          return
+          insertComponentStyles(childComponent, node.package ?? package_name)
```
jacob-kofoed, line 160: This change improves styles update time on eg. EditorPage on the toddle project from 250ms to about 40ms.

Please double check the logic, but to me it looks the same as we are recursively going through elements, before we just did it many more times than needed.

packages/runtime/src/editor-preview.main.ts
```diff
@@ -1372,39 +1421,93 @@ export const createRoot = (
...
+            )
+          })
+          .forEach((nodeId) => {
+            const node = newCtx.component.nodes[nodeId] as
+              | ComponentNodeModel
+              | ElementNodeModel
+            const oldNode = ctx?.component.nodes[nodeId] as
+              | ComponentNodeModel
+              | ElementNodeModel
+            const prevClassName = getClassName([
```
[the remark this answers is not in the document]
jacob-kofoed: No, the logic should force the correct types. I'll see if I can come up with a better way to pass the object through the functions rather than the ID to enforce types.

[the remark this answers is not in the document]
jacob-kofoed: I'm just using the forEach to keep types rather than filter then forEach.

PR #209, opened 2025-03-31
Fixed issue where meta key could be in pressed state on click

ChrEsb:
This could happen when you tab out of a window and back. The editor never registers the "keyup" event and thus the cmd key appears pressed.

This might not be a proper fix. We set the value of the metaKey on keydown and keyup and since keyup isnt triggered when tabbing back into the window we might have other issues with this? I assume storing the meta key value is used for drag n' drop?

But since click and dblclick needs the exact current value I now read the value of the meta key from the event and not the stored value. Let me know what you think.


jacob-kofoed (approved):
This is probably the way we want to handle keys going forward ✅

PR #210, opened 2025-03-31
Fix an issue where instance styling would not be applied in the correct order

jacob-kofoed:
Order is very important for editor styling. So we have to apply it all each time or otherwise create some complex order management.

This is still a lot faster than before as we apply styles only ones instead of nearly O(n^2) as previously. I utilize fragments to ensure no relayouts between adding styles (most browsers likely does this optimization anyway).

### DIFF
--- a/packages/runtime/src/editor-preview.main.ts
+++ b/packages/runtime/src/editor-preview.main.ts
@@ -372,12 +372,6 @@ export const createRoot = (
             return
           }
           if (message.data.component.name != component?.name) {
-            // Clean up all styles before switching components
-            document.head
-              .querySelectorAll('style[data-hash]')
-              .forEach((style) => {
-                style.remove()
-              })
             showSignal.cleanSubscribers()
           }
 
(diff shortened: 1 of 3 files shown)


packages/runtime/src/editor-preview.main.ts
```diff
@@ -372,12 +372,6 @@ export const createRoot = (
             return
           }
           if (message.data.component.name != component?.name) {
-            // Clean up all styles before switching components
```
jacob-kofoed, line 375: No need for this anymore as styles is cleaned in the updateStyles function.

packages/runtime/src/editor-preview.main.ts
```diff
@@ -1162,15 +1156,15 @@ export const createRoot = (
           }
         }
       }
-      const selectedNode = getDOMNodeFromNodeId(selectedNodeId)
```
jacob-kofoed, line 1165: Moving this outside the condition as the rect was not updated when going from a variant to the default.

packages/runtime/src/styles/style.ts
```diff
@@ -158,19 +162,35 @@ ${
...
-      }
-      parent.appendChild(getNodeStyles(node, classHash))
     })
   }
+
   insertComponentStyles(root, undefined)
+
+  // Remove old styles.
+  // We do not keep track of changes, so must remove all and re-add as order matters.
+  parent.querySelectorAll('[data-hash]').forEach((node) => node.remove())
```
jacob-kofoed, line 186: Feels bad to redraw everything even when we often just update a single rule. I may revisit this to see if we can do some efficient diff.

erik-beus (approved):

PR #218, opened 2025-04-02
Add rules for unknown formula inputs

erik-beus:
These rules will find path formulas that point to unknown inputs/args in component/project formulas

### DIFF
--- a/packages/search/src/rules/formulas/unknownProjectFormulaInputRule.ts
+++ b/packages/search/src/rules/formulas/unknownProjectFormulaInputRule.ts
@@ -20,7 +20,6 @@ export const unknownProjectFormulaInputRule: Rule<{
     ) {
       return
     }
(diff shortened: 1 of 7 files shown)


packages/search/src/rules/formulas/unknownProjectFormulaInputRule.ts
```diff
@@ -0,0 +1,32 @@
...
+      nodeType !== 'formula' ||
+      value.type !== 'path' ||
+      value.path?.[0] !== 'Args' ||
+      value.path.length < 2 ||
+      path[0] !== 'formulas' ||
+      path.length < 2
+    ) {
+      return
+    }
+    console.log({ path, value })
```
jacob-kofoed, line 23: No logging plx

jacob-kofoed (approved):
Looks good

PR #219, opened 2025-04-02
Add rule for unknown repeat item formulas

erik-beus:
This rule will detect unknown path formulas to repeat items in component nodes

### DIFF
--- a/packages/search/src/rules/formulas/unknownRepeatItemFormulaRule.ts
+++ b/packages/search/src/rules/formulas/unknownRepeatItemFormulaRule.ts
@@ -1,0 +1,48 @@
+import type { NodeModel } from '@toddledev/core/dist/component/component.types'
+import type { Rule } from '../../types'
+
+export const unknownRepeatItemFormulaRule: Rule = {
+  code: 'unknown repeat item formula',
+  level: 'error',
+  category: 'Unknown Reference',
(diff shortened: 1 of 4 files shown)


packages/search/src/rules/formulas/unknownRepeatItemFormulaRule.ts
```diff
@@ -0,0 +1,48 @@
...
+      value.type !== 'path' ||
+      value.path?.[0] !== 'ListItem' ||
+      value.path?.[1] !== 'Item' ||
+      path.length < 3 ||
+      path[0] !== 'components' ||
+      path[2] !== 'nodes'
+    ) {
+      return
+    }
+    console.log({ path, value })
```
jacob-kofoed, line 20: We need an issue rule to detect issues in our issue rules, like this console.log issue in this issue rule.

packages/search/src/rules/formulas/unknownRepeatItemFormulaRule.ts
```diff
@@ -0,0 +1,48 @@
...
+    const [_components, componentName, _nodes, nodeId] = path as string[]
+    const component = files.components[componentName]
+    console.log({ component })
```
jacob-kofoed, line 23: This console.log as well is an issue

jacob-kofoed (approved):

PR #223, opened 2025-04-03
Fixed an issue where url params was available in formula editor on a …

ChrEsb:
…component

This could be reproduced by doing the following:

1. Navigate to a page
2. Navigate to a component
3. Open the formula editor.
4. Open the formula selector

You will see that URL params are available from the previous Page.

### DIFF
--- a/packages/runtime/src/editor-preview.main.ts
+++ b/packages/runtime/src/editor-preview.main.ts
@@ -1355,10 +1355,12 @@ export const createRoot = (
     dataSignal.update((data) => {
       return {
         ...data,
-        'URL parameters': {
-          ...window.toddle.locationSignal.get().query,
-          ...window.toddle.locationSignal.get().params,
-        } as Record<string, string>,
+        'URL parameters': component?.route
+          ? ({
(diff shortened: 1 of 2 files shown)


jacob-kofoed (approved):
Change makes sense, seems to work as well ✅

packages/runtime/src/editor-preview.main.ts
```diff
@@ -1355,10 +1355,12 @@ export const createRoot = (
     dataSignal.update((data) => {
       return {
         ...data,
-        'URL parameters': {
-          ...window.toddle.locationSignal.get().query,
-          ...window.toddle.locationSignal.get().params,
-        } as Record<string, string>,
+        'URL parameters': component?.route
```
erik-beus, line 1358: In theory you should also check for the old definition of a page (a path property perhaps?), but we haven't seen any of those pages for a while I believe, so this is fine
Perhaps you could use the utility function though: isPageComponent

erik-beus (changes requested):

erik-beus (approved):

PR #225, opened 2025-04-04
Add unknown repeat index formula rule

erik-beus:
Repeat index formulas are different from repeat item formulas, and I suggest we use separate rules for them.

### DIFF
--- a/packages/search/src/rules/formulas/unknownRepeatIndexFormulaRule.test.ts
+++ b/packages/search/src/rules/formulas/unknownRepeatIndexFormulaRule.test.ts
@@ -1,0 +1,138 @@
+import { searchProject } from '../../searchProject'
+import { unknownRepeatIndexFormulaRule } from './unknownRepeatIndexFormulaRule'
+
+describe('unknownRepeatIndexFormulaRule', () => {
+  test('should find unknown repeat index formulas', () => {
+    const problems = Array.from(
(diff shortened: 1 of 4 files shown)


packages/search/src/rules/formulas/unknownRepeatIndexFormulaRule.test.ts
```diff
@@ -0,0 +1,138 @@
...
+                    type: 'path',
+                    path: ['ListItem', 'Index'],
+                  },
+                },
+                'Z95Ucsbip-YWbTmC38-vG': {
+                  tag: 'li',
+                  type: 'element',
+                  attrs: {},
+                  style: {
+                    'font-weight': 'var(--font-weight-regular)',
```
jacob-kofoed, line 26: Nice of you to add styles to our tests

jacob-kofoed (approved):
Looks good ✅ I like how we have test coverage on all the issue rules 😄

packages/search/src/rules/formulas/unknownRepeatIndexFormulaRule.test.ts
erik-beus: Very important 🙈 I just copied something from a node in my project. That's usually a quick way of building up the structure for a test 👍

PR #228, opened 2025-04-04
Adjust unknown input rules to ignore parent/item

erik-beus:
Since path formulas for a formula's item or a parent formula are also using the `Args` scope, we would report more issues than necessary.


jacob-kofoed (approved):
Works on my example. Dunno if you can get access to https://discord.com/channels/972416966683926538/1357652579630186556/1357652579630186556 to test, but better just get it out and ask later.

PR #253, opened 2025-04-24
Continiously report selection and highlight rect to better support animations

jacob-kofoed:
Performance-wise, this is < 1ms, but it does run every frame. It feels a bit wrong, but I think it is the cleanest and most easily maintained way to make animated elements show nicely in the editor.


packages/runtime/src/editor-preview.main.ts
```diff
@@ -1661,6 +1569,58 @@ export const createRoot = (
       testMode: mode === 'test',
     })
   }
+
+  // Animations are first class citizens in Nordcraft, so we sync their overlay positions on each frame
```
jacob-kofoed, line 1570: Dunno if you agree with this statement, but I don't see a much better way if we really want to support animated elements fully.

packages/runtime/src/editor/drag-drop/dragStarted.ts
```diff
@@ -99,14 +99,14 @@ export function dragStarted({
...
-    dragState.repeatedNodes.forEach((node, i) => {
+    dragState.repeatedNodes.forEach((node) => {
       // Calculate rect without rotation as it expands the rect and makes it difficult to calculate the correct position
       node.style.setProperty('rotate', '0deg')
       const fromRect = node.getBoundingClientRect()
       node.style.removeProperty('rotate')
       const toX = followRect.left + followRect.width / 2 - fromRect.width / 2
       const toY = followRect.top + followRect.height / 2 - fromRect.height / 2
-      const interpolation = 0.4 / (i + 1)
+      const interpolation = 0.4
```
jacob-kofoed, line 109: Andreas asked to have repeated nodes follow the cursor more tightly when drag/dropping

erik-beus (approved):

PR #259, opened 2025-04-29
Add rule for detecting actions after navigate

erik-beus:
Performing any action after using a "Go to URL" action is error prone and there's a chance those actions will not be executed.
This rule detects any action that is executed after a "Go to URL" action in:

- Events
- Workflows
- On load events
  
Closes https://github.com/nordcraftengine/nordcraft-internal/issues/2007

### DIFF
--- a/packages/search/src/rules/workflows/noPostNavigateAction.ts
+++ b/packages/search/src/rules/workflows/noPostNavigateAction.ts
@@ -23,8 +23,9 @@ export const noPostNavigateAction: Rule<{ parameter: string }> = {
       return
     }
     const actionIndex = Number(_actionIndex)
-    actions.slice(actionIndex + 1).forEach((_, i) => {
-      report([...actionsArrayPath, String(actionIndex + 1 + i)])
-    })
+    if (actionIndex < actions.length - 1) {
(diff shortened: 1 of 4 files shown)


packages/search/src/rules/workflows/noPostNavigateAction.ts
```diff
@@ -0,0 +1,30 @@
...
+  visit: (report, { nodeType, path, value, files }) => {
+    if (
+      nodeType !== 'action-model' ||
+      value.type !== undefined ||
+      value.name !== '@toddle/gotToURL'
+    ) {
+      return
+    }
+    const actionsArrayPath = path.slice(0, -1).map((p) => String(p))
+    const actions = get(files, actionsArrayPath)
```
jacob-kofoed, line 17: Nice idea to use `get` 🌮

jacob-kofoed (approved):
I love it!

I'm not sure how it works, or how exactly it should work if a navigate is inside a switch case, but maybe it should also report those.

You could also consider reversing the rule to not spam too many issues, by reporting the navigate function if it is not the last instead of reporting subsequent actions. Again, I am not sure what is best.

PR #267, opened 2025-05-01
Treat packages as partial when accessing them

erik-beus:
I added an optional chaining + adjusted the type for packages

I also moved a utility function to the ssr package


marija04 (approved):
Looks good 😊

jacob-kofoed (approved):

PR #279, opened 2025-05-07
Force the cursor to always be within the dragged element

jacob-kofoed:
Small improvement to drag and drop. Sometimes when dragging an element, and that element would greatly change the layout of the page, the dragging-point could move outside the element. This change makes it so the dragged element will move towards the cursor in such events.


marija04 (approved):
Looks good!

PR #287, opened 2025-05-08
Update references copy

whitep4nth3r:
This PR addresses some inconsistencies and adds/removes some things that were breaking the docs displays, most notably the links to MDN which produced this (I think this was the culprit):

image

### DIFF
--- a/packages/lib/formulas/concatenate/formula.json
+++ b/packages/lib/formulas/concatenate/formula.json
@@ -12,6 +12,17 @@
           { "type": "Array", "ofType": { "type": "Any" } }
         ]
       }
+    },
+     {
+      "name": "1",
+      "formula": { "type": "value", "value": "" },
+      "type": {
+        "type": "Union",
+        "types": [
+          { "type": "String" },
+          { "type": "Array", "ofType": { "type": "Any" } }
+        ]
+      }
     }
   ],
   "variableArguments": true,
--- a/packages/lib/formulas/concatenate/formula.json
+++ b/packages/lib/formulas/concatenate/formula.json
@@ -12,17 +12,6 @@
           { "type": "Array", "ofType": { "type": "Any" } }
         ]
       }
-    },
-     {
-      "name": "1",
-      "formula": { "type": "value", "value": "" },
-      "type": {
-        "type": "Union",
-        "types": [
-          { "type": "String" },
-          { "type": "Array", "ofType": { "type": "Any" } }
-        ]
-      }
     }
   ],
   "variableArguments": true,
--- a/packages/lib/formulas/formatDate/formula.json
+++ b/packages/lib/formulas/formatDate/formula.json
@@ -15,12 +15,12 @@
     },
     {
       "name": "Options",
-      "description": "Optional Object for configuring the formatting of the output. See https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/DateTimeFormat/DateTimeFormat#syntax for all options.",
+      "description": "Optional Object for configuring the formatting of the output.",
       "type": { "type": "Object" },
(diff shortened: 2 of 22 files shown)


packages/lib/formulas/concatenate/formula.json
```diff
@@ -12,6 +12,17 @@
...
+     {
+      "name": "1",
+      "formula": { "type": "value", "value": "" },
+      "type": {
+        "type": "Union",
+        "types": [
+          { "type": "String" },
+          { "type": "Array", "ofType": { "type": "Any" } }
+        ]
+      }
```
erik-beus, line 25: Technically, it could also receive just 1 argument. I don't think we should necessarily change this, but instead use the `variableArguments` flag to somehow indicate in the docs that you can pass in multiple parameters.
```suggestion
```

packages/lib/formulas/formatDate/formula.json
```diff
@@ -15,12 +15,12 @@
     },
     {
       "name": "Options",
-      "description": "Optional Object for configuring the formatting of the output. See https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/DateTimeFormat/DateTimeFormat#syntax for all options.",
+      "description": "Optional Object for configuring the formatting of the output.",
```
erik-beus, line 18: I think the links are relevant. But let's render them as markdown instead and make sure we do the same in the editor when we show their descriptions?
```suggestion
      "description": "Optional Object for configuring the formatting of the output. See [MDN](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/DateTimeFormat/DateTimeFormat#syntax) for all options",
```

packages/lib/formulas/formatDate/formula.json
```diff
@@ -15,12 +15,12 @@
...
       "type": { "type": "Object" },
       "formula": { "type": "value", "value": null }
     }
   ],
-  "description": "Format a date using the Intl.DateTimeFormat API. See https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/DateTimeFormat/DateTimeFormat for more information.",
+  "description": "Format a date using the Intl.DateTimeFormat API.",
```
erik-beus, line 23: Same as above

packages/lib/formulas/formatNumber/formula.json
```diff
@@ -15,12 +15,12 @@
     },
     {
       "name": "Options",
-      "description": "Optional Object for configuring the formatting of the output. See https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/NumberFormat/NumberFormat#locale_options for all options.",
+      "description": "Optional Object for configuring the formatting of the output.",
```
erik-beus, line 18: Same

packages/lib/formulas/formatNumber/formula.json
```diff
@@ -15,12 +15,12 @@
...
       "type": { "type": "Object" },
       "formula": { "type": "value", "value": null }
     }
   ],
-  "description": "Format a Number using the Intl.NumberFormat API. See https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/NumberFormat for more information.",
+  "description": "Format a Number using the Intl.NumberFormat API.",
```
erik-beus, line 23: Same

erik-beus (changes requested):
Please see suggestions

packages/lib/formulas/formatDate/formula.json
whitep4nth3r: Yes please.

erik-beus (approved):

PR #313, opened 2025-05-12
Adjust default request body content type

erik-beus:
Currently, we only support FormData, URL encoding, JSON, and text as content-type for API requests. However, if a different content-type has been specified, we should not try to json encode the request body.
In order to make this a smaller breaking change than it could be, we should still default to json encoding the request body if no content-type is specified (when it's null). If it's specified and its none of the above though, we should not try to encode the request body.


ChrEsb (approved):
This is a better approach imo

cullophid:
Yes this makes sense. I think it should be safe to release.

Im wondering if we should warn people ahead of time
Just to be sure

PR #316, opened 2025-05-13
Add script for sending release notes to Discord

erik-beus:


jacob-kofoed (approved):
I didn't test it, but I assume you tested the hook in Make. Excited about what else we can do with Octokit 🤔 Maybe automatic release notes, etc.

PR #317, opened 2025-05-13
Fixed an issue with even-child selector not working as expected

jacob-kofoed:
Fixed an issue where using the `even-child` selector in the editor would not apply the selector.

I tested on this branch https://tester-jacob_test.toddle.site/repeat (each even card should have a lighter background color).

### DIFF
--- a/packages/core/src/styling/variantSelector.ts
+++ b/packages/core/src/styling/variantSelector.ts
@@ -38,7 +38,7 @@ type MediaQuery = {
 
 export interface StyleVariant {
   autofill?: boolean
-  'even-child'?: boolean
+  'nth-child(even)'?: boolean
   'first-child'?: boolean
   'first-of-type'?: boolean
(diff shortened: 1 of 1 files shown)


packages/core/src/styling/variantSelector.ts
```diff
@@ -38,7 +38,7 @@ type MediaQuery = {
 
 export interface StyleVariant {
   autofill?: boolean
-  'even-child'?: boolean
+  'nth-child(even)'?: boolean
```
jacob-kofoed, line 41: I don't believe this is a breaking change as we've always saved it as `nth-child(even)`, so it never worked 🤦

jacob-kofoed: Or actually, I don't know that for sure, but I assume we never saved by `even-child`. Little scary perhaps.

PR #318, opened 2025-05-13
Only render animations with the same key once

jacob-kofoed:
Animations are stored with their hashed value, so rendering duplicates doesn't make sense. Duplicate `@keyframes` is not an issue AFAIK, but this way we save a tiny bit of bandwidth after GZIP and a lot before.


packages/core/src/styling/hash.ts
```diff
@@ -0,0 +1,46 @@
...
+
+  while (i) {
+    h = (h * 33) ^ x.charCodeAt(--i)
+  }
+
+  return h
+}
+
+// This is a djb2 hashing function
+export const hash = (x: string) => {
```
jacob-kofoed, line 18: Random refactor. I just didn't think this belonged in a file called `className`.

packages/core/src/styling/style.css.ts
```diff
@@ -211,6 +212,11 @@ ${selector}::-webkit-scrollbar {
           node.animations
             ? Object.entries(node.animations)
                 .map(([animationName, keyframes]) => {
+                  // Animation names are stored by their hash, so no need to render them more than once.
+                  if (animationHashes.has(animationName)) {
```
jacob-kofoed, line 216: We do something similar with class hashes.

erik-beus (approved):

PR #319, opened 2025-05-13
Omit browser bug and improve performance by not transitioning out of bounds elements

jacob-kofoed:
There might be some paint crashing issues in Chromium browsers around view transition of elements inside an iframe, but out of bounds of their viewport. This change should improve performance for extreme cases either way performance.

### DIFF
--- a/packages/runtime/src/editor/drag-drop/dragEnded.ts
+++ b/packages/runtime/src/editor/drag-drop/dragEnded.ts
@@ -1,3 +1,4 @@
+import { isElementInViewport } from '../../utils/isElementInViewport'
 import { tryStartViewTransition } from '../../utils/tryStartViewTransition'
 import type { DragState } from '../types'
 import { DRAG_MOVE_CLASSNAME } from './dragMove'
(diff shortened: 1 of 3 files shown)


packages/runtime/src/editor/drag-drop/dragEnded.ts
```diff
@@ -18,14 +19,14 @@ export async function dragEnded(dragState: DragState, canceled: boolean) {
...
-    if (node instanceof HTMLElement) {
+    if (node instanceof HTMLElement && isElementInViewport(node)) {
       node.style.setProperty(
         'view-transition-name',
         'dropped-item-sibling-' + i,
       )
     }
   })
-  dragState.repeatedNodes.forEach((node, i) => {
+  dragState.repeatedNodes.filter(isElementInViewport).forEach((node, i) => {
```
jacob-kofoed, line 29: Can be tested to see that there is still some smooth transitions when dropping an element after drag and drop.

packages/runtime/src/utils/isElementInViewport.ts
```diff
@@ -0,0 +1,14 @@
+export function isElementInViewport(element: HTMLElement, offset = 0): boolean {
```
erik-beus, line 1: Consider renaming `offset`
```suggestion
export function isElementInViewport(element: HTMLElement, margin = 0): boolean {
```
```suggestion
export function isElementInViewport(element: HTMLElement, threshold = 0): boolean {
```

erik-beus (approved):
Looks good. Just 1 suggestion

PR #328, opened 2025-05-14
Added label as optional field on formulas to sync with how theyre use…

ChrEsb:
…d in the editor

### DIFF
--- a/package.json
+++ b/package.json
@@ -2,7 +2,7 @@
   "name": "nordcraft",
   "type": "module",
   "license": "Apache-2.0",
-  "version": "1.0.10",
+  "version": "1.0.11",
   "homepage": "https://github.com/nordcraftengine/nordcraft",
   "private": "false",
   "workspaces": [
--- a/packages/core/src/formula/formula.ts
+++ b/packages/core/src/formula/formula.ts
@@ -4,8 +4,8 @@ import type {
   CustomFormulaHandler,
   FormulaHandler,
   FormulaLookup,
+  NordcraftMetadata,
   Toddle,
-  ToddleMetadata,
 } from '../types'
 import { isDefined, toBoolean } from '../utils/util'
 import { isToddleFormula } from './formulaTypes'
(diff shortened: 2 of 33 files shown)


packages/core/src/formula/formula.ts
```diff
@@ -14,7 +14,22 @@ import { isToddleFormula } from './formulaTypes'
...
+    | 'path'
+    | 'function'
+    | 'object'
+    | 'array'
+    | 'record'
+    | 'value'
+    | 'apply'
+    | 'or'
+    | 'and'
+    | 'switch'
```
erik-beus, line 28: These are all added in the individual types below and allow us to work with discriminated unions. Why should they also be added here?
```suggestion
```

packages/core/src/formula/formula.ts
```diff
@@ -14,7 +14,22 @@ import { isToddleFormula } from './formulaTypes'
 declare const document: Document | undefined
 type ShadowRoot = DocumentFragment
 
-export interface PathOperation extends ToddleMetadata {
+interface BaseOperation extends ToddleMetadata {
```
erik-beus, line 17: ;)
```suggestion
interface BaseOperation extends NordcraftMetadata {
```

package.json
```diff
@@ -2,7 +2,7 @@
   "name": "nordcraft",
   "type": "module",
   "license": "Apache-2.0",
-  "version": "1.0.10",
+  "version": "1.0.11",
```
erik-beus, line 5: Is it important enough to bump the version for this change?

erik-beus (changes requested):
Please see comments

package.json
ChrEsb: This is a fix I need for my current task in our internal project. So bumping the version will allow me to continue and merge that work without waiting for enough changes to be accumulated for us to bump the version.

erik-beus (approved):

PR #345, opened 2025-05-15
Allow enabling/disabling routes

erik-beus:
There are scenarios where you only want a route (redirect or rewrite) to be enabled in certain scenarios:
- If a cookie is declared
- If a user prefers a certain language
- If you're on a specific origin

I suggest we add an `enabled` flag for routes to control this (defaults to `true` for existing routes).
This might be more relevant for redirects than rewrites, but I don't see any drawback in allowing rewrites to be optional as well.

This means:
- You can declare an optional redirect/rewrite that can have an identical path to a page. This will allow you to serve a page when the redirect/rewrite isn't applied.
- You could technically have multiple routes for the same source declaration if the `enabled` flag was different for each route
- We would not necessarily know which route to pick if multiple routes existed for the same source, but would probably just pick the first one that was enabled

### DIFF
--- a/package.json
+++ b/package.json
@@ -2,7 +2,7 @@
   "name": "nordcraft",
   "type": "module",
   "license": "Apache-2.0",
-  "version": "1.0.22",
+  "version": "1.0.23",
   "homepage": "https://github.com/nordcraftengine/nordcraft",
   "private": "false",
   "workspaces": [


packages/backend/src/middleware/routeLoader.ts
```diff
@@ -15,8 +19,17 @@ export const routeLoader = createMiddleware<HonoEnv<HonoRoutes & HonoRoute>>(
...
+      env: serverEnv({
+        branchName: 'main',
+        req: ctx.req.raw,
+        logErrors: false,
+      }),
+      req: ctx.req.raw,
       url,
-      routes: ctx.var.routes?.routes ?? {},
+      routes: ctx.var.routes.routes,
+      // TODO: We should pass in global toddle formulas from project + packages here
```
ChrEsb, line 31:
```suggestion
      // TODO: We should pass in global Nordcraft formulas from project + packages here
```

packages/backend/src/routes/routeHandler.ts
```diff
@@ -11,12 +14,12 @@ export const routeHandler = async (
...
   const destination = getRouteDestination({
-    files: {} as any, // c.var.project.files,
-    req: c.req.raw,
-    route,
-
     // might not want to use main branch here
     env: serverEnv({ branchName: 'main', req: c.req.raw, logErrors: false }),
+    req: c.req.raw,
+    route,
+    // TODO: We should pass in global toddle formulas from project + packages here
```
ChrEsb, line 21:
```suggestion
    // TODO: We should pass in global Nordcraft formulas from project + packages here
```

ChrEsb (changes requested):

ChrEsb (approved):

PR #387, opened 2025-05-28
Allow template values in body of proxied requests

erik-beus:
To support use cases like refreshing access tokens, we must allow injecting cookie template values in the body of proxied requests. This will be an opt in, and we will only process the body during proxying when the special header is set/available.

- Add tests for proxied API requests
- Rename header for destination URL for proxied requests
- Make sure we run backend tests in CI

### DIFF
--- a/packages/backend/src/routes/apiProxy.test.ts
+++ b/packages/backend/src/routes/apiProxy.test.ts
@@ -78,9 +78,8 @@ describe('API proxy', () => {
         targetUrl.searchParams.get('param2'),
       )
       expect(req.method).toBe('POST')
-      console.log('Incoming request', req)
       const body = await req.text()
-      expect(body).toBe('hello world') // The body should not be modified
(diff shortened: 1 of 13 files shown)


packages/backend/src/routes/apiProxy.test.ts
```diff
@@ -0,0 +1,169 @@
...
+        init: {
+          body: JSON.stringify({
+            myToken: STRING_TEMPLATE('cookies', 'refresh_token'),
+          }),
+        },
+      },
+    )
+    expect(res.status).toBe(200)
+  })
+  it("Should ignore the request's body if the special", async () => {
```
ChrEsb: This sentence seems to abruptly end?

ChrEsb (approved):
Approved with a comment

packages/backend/src/routes/apiProxy.test.ts
erik-beus: It was a test to see if you were awake 😂

PR #391, opened 2025-06-02
Respect duplicate headers in API responses

erik-beus:
Proxied API requests would loose information about duplicate headers (multiple `set-cookie` headers for instance). This change ensures that all headers are preserved and accessible in the API response.

It also introduces a utility function to merge duplicate headers (by comma separating them) when needed, which is useful for headers like `set-cookie`.

This issue was reported (for proxied API requests) here https://discord.com/channels/972416966683926538/1378707406363824128

### DIFF
--- a/packages/backend/src/routes/apiProxy.ts
+++ b/packages/backend/src/routes/apiProxy.ts
@@ -81,7 +81,7 @@ export const proxyRequestHandler = async (
 
     const returnResponse = new Response(body, {
       status: response.status,
-      headers: Object.fromEntries(response.headers),
+      headers: new Headers(response.headers),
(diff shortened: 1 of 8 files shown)


packages/backend/src/routes/apiProxy.ts
```diff
@@ -81,7 +81,7 @@ export const proxyRequestHandler = async (
 
     const returnResponse = new Response(body, {
       status: response.status,
-      headers: Object.fromEntries(response.headers),
+      headers: new Headers(response.headers),
```
erik-beus, line 84: This solves the reported issue

PR #428, opened 2025-06-11
Correctly handle type attribute on svgs

ChrEsb:


packages/runtime/src/utils/setAttribute.ts
```diff
@@ -19,16 +19,22 @@ export function setAttribute(
...
-      if (elem instanceof HTMLProgressElement) {
-        // An HTMLProgressElement will break other elements in our editor if the value is not a (finite) number
-        // See docs here https://developer.mozilla.org/en-US/docs/Web/HTML/Element/progress#value
-        // and original issue here https://discord.com/channels/972416966683926538/1317827591230722048
-        if (!isDefined(value) || !Number.isFinite(Number(value))) {
-          val = 0
+      if (elem instanceof SVGElement) {
+        // The type attribute on SVG elements should just be added as an attribute
+        // See https://developer.mozilla.org/en-US/docs/Web/SVG/Attribute/type
+        elem.setAttribute(attr, String(value))
```
erik-beus: I suppose we still don't want to set `null`/`undefined` here?
```suggestion
        if (toBoolean(val)) {
          elem.setAttribute(attr, String(value))
        }
```

ChrEsb: Good point. 👍

PR #443, opened 2025-06-17
feat: add popular tag to elements

mikkelnordcraft:
For sorting on the frontend.


packages/editor/elements/_dev_buildElements.ts
```diff
@@ -69,6 +90,7 @@ const init = () => {
           `An SVG element representing the SVG ${element} element.`,
         link: `https://developer.mozilla.org/en-US/docs/Web/SVG/Reference/Element/${element}`,
         aliases: aliases,
+        isPopular: popularElements.includes(element) ? true : undefined,
```
erik-beus: There can be name clashes between elements and svg elements. I suggest you introduce a separate array of `popularSvgElements` to make sure we don't have name clashes. For instance there's a `title` element in html and for svgs (although we don't include the `title` html element atm)

erik-beus (changes requested):

packages/editor/elements/_dev_buildElements.ts
mikkelnordcraft: 🫡

PR #459, opened 2025-06-26
Report empty/nullish required attributes

erik-beus:
Currently this rule is used for:

- img -> alt
- img -> src
- a -> href

All of the above would be considered illegal if they are empty or nullish. These updates will ensure that empty/nullish attributes are also reported as errors for the `createRequiredElementAttributeRule`.


erik-beus:
> I am pretty sure empty alt tags is valid, in cases where screenreaders should skip the image as being decorative. The important part is to have the alt tag, but giving it an empty value is valid.

Good point. I've updated the rule and the default value of the `alt` attribute on elements for our element catalog 👍

jacob-kofoed (approved):

PR #462, opened 2025-06-26
End to end testing with playwright

jacob-kofoed:
Proof of concept for first nordcraft e2e tests.

We can use 2e2 to catch regressions on a list of user-sites and the editor itself. We can also use it as a form of specs while developing new features.

Preferably, this should run on a localhost on each Github PR, and potentially run the editor specific tests on Nordcraft PRs eventually.

### DIFF
--- a/e2e/tests/editor-api.spec.ts
+++ b/e2e/tests/editor-api.spec.ts
@@ -23,7 +23,8 @@ test('can create, name, and delete an API', async ({ page }) => {
   expect(page.getByText('My Renamed API')).toBeDefined()
 
   // Delete the API
-  await apiElement.click({ button: 'right' })
+  const apiElementNew = page.getByTestId('api-item')
+  await apiElementNew.click({ button: 'right' })
   await page.getByText('Delete').click()
   await expect(page.getByTestId('api-item')).toHaveCount(0)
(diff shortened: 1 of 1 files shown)


packages/e2e/tests/Editor-style-variable.spec.ts
```diff
@@ -0,0 +1,26 @@
...
+
+  // set the viewport size
+  await page.setViewportSize({ width: 1920, height: 1080 });
+});
+
+test("can change standard style properties", async ({ page }) => {
+  await page.goto("/projects/toddle_e2e/branches/main/components/HomePage?rightpanel=style&selection=nodes.wz0B5AQwNmG5sc0BSRfvD");
+
+  const canvasElement = await page.frameLocator("#toddle-canvas").getByTestId("0.0.0");
+  const stylePanelWidthInput = page.getByTestId("0.0[default].1.0[default].0[default].0[default].1[default].0.3.0.1.2.1.0.0.0.0.1.0[Content].0.0.0.0[default].0[default].0[default].2");
```
jacob-kofoed: We should probably change our test-id to actually use `data-test-id` and set those specifically on the elements we want to test.

PR #474, opened 2025-07-04
Experimental preview backend that relies on DO

erik-beus:
Add a different version of the backend that relies on a project's state from our internal Durable Object. This allows us to render/show the latest changes on a branch.


packages/backend/preview.wrangler.toml
```diff
@@ -0,0 +1,20 @@
...
+bindings = [
+  { name = "BRANCH_STATE", class_name = "BranchController", script_name = "worker" }
+]
+
+[assets]
+directory = "./assets"
+
+[vars]
+PROJECT_SHORT_ID = "awesome_blog"
+BRANCH_NAME = "a"
```
ChrEsb: These seem rather project specific. Are they needed?

erik-beus: Lol, they are. We could just pass in env vars when we start the preview locally I suppose. I will adjust it 👍

PR #477, opened 2025-07-07
Handle style-properties v2

jacob-kofoed:
Style properties v2 allows for some new features that the old structure would not:

- Properties are declared globally in CSS with `@property`. This allows for transitions, animations and better calc.
- Properties are defined in a constructed style-sheet rather than using the styles attribute. This frees up the styles-attribute for users.
- By not using the styles attribute, we now allow instance styling of components and overriding style variables on style variants.

I will be cleaning up this PR up a bit more.


packages/runtime/src/styles/StylePropertyStyleSheet.ts
```diff
@@ -0,0 +1,171 @@
...
+  public getStyleSheet(): CSSStyleSheet {
+    return this.styleSheet
+  }
+
+  /**
+   * Maps all selectors to their rule index. This is used to map the initial SSR style variable values
+   * to their selectors.
+   * This is only called once, when the first property is registered.
+   */
+  private generateRuleMap() {
```
jacob-kofoed: This is probably the funkiest part of the PR. It seems to work, but is probably quite brittle. Alternative is the classic hydration strategy of just tearing down the SSR content (style-sheet in this case) and then build a new one. I am attempting to resume from SSR by mapping selectors.

packages/core/src/component/component.types.ts
```diff
@@ -59,7 +60,9 @@ export interface StyleVariant {
   mediaQuery?: MediaQuery
   breakpoint: 'small' | 'medium' | 'large'
   startingStyle?: boolean
+  pseudoElement?: string
   style: NodeStyleModel
+  'style-variables'?: Array<StyleVariable>
```
erik-beus: Should `style-variables` perhaps be an object and not an array? That would follow our general principle, and you could get rid of the `name` property?

```suggestion
  'style-variables'?: Partial<Record<string, StyleVariable>>
```

packages/core/src/component/component.types.ts
```diff
@@ -90,12 +108,7 @@ export interface ElementNodeModel {
   children: string[]
   events: Record<string, EventModel>
   classes: Record<string, { formula?: Formula }>
-  'style-variables'?: Array<{
-    category: StyleTokenCategory
-    name: string
-    formula: Formula
-    unit?: string
-  }>
+  'style-variables'?: Array<StyleVariableLegacy | StyleVariable>
```
erik-beus: Changing this into an object is more tricky. Perhaps we would need a different naming?

```suggestion
  'style-variables'?: Array<StyleVariableLegacy>
  'style-variables-final'?: Partial<Record<string, StyleVariable>>
```

packages/core/src/styling/style.css.ts
```diff
@@ -235,6 +240,40 @@ ${selector}::-webkit-scrollbar {
...
+            )
+            if (existingVariable) {
+              // Warn if the style variable is already registered with a different syntax
+              if (
+                existingVariable.type === 'primitive' &&
+                styleVariable.syntax.type === 'primitive' &&
+                existingVariable.name !== styleVariable.syntax.name
+              ) {
+                // eslint-disable-next-line no-console
+                console.warn(
```
erik-beus, line 260: It's nice to include this check. Would it be possible to add a test that covers this scenario?

packages/core/src/styling/styleProperty.ts
```diff
@@ -0,0 +1,81 @@
+import type { Formula } from '../formula/formula'
+
+type CssSyntaxPrimitive =
```
erik-beus: A link to MDN here would be helpful 🙏 It seems we're only adding a subset of properties? https://developer.mozilla.org/en-US/docs/Web/CSS/length

packages/core/src/styling/styleProperty.ts
```diff
@@ -0,0 +1,81 @@
...
+  | 'number'
+  | 'percentage'
+  | 'resolution'
+  | 'string'
+  | 'time'
+  | 'transform-function'
+  | 'transform-list'
+  | 'url'
+
+export type CssSyntax = CssSyntaxPrimitive
```
erik-beus: Why do you have a duplicate type here?

packages/core/src/styling/styleProperty.ts
```diff
@@ -0,0 +1,81 @@
...
+  syntax: CssSyntaxNode // The syntax definition for the property
+  formula: Formula // The value of the property. The formula may be a static string value to point to another property, or a more complex dynamical value.
+}
+
+/**
+ * Default initial values for CSS properties based on their syntax type.
+ *
+ * Note: Eventually, the initial values may be set in a theme or somewhere global.
+ */
+const DEFAULT_INITIAL_VALUE_BY_SYNTAX: Record<CssSyntax, string | number> = {
```
erik-beus: This seems more accurate?
```suggestion
const DEFAULT_INITIAL_VALUE_BY_SYNTAX: Record<CssSyntax, string> = {
```

Also, I believe Typescript wouldn't complain atm if you leave out a property. If you change the `CssSyntax` to be an enum, I believe Typescript would complain if you used `Record<typeof keyof CssSyntax, string>`?

packages/runtime/src/components/STYLE_VARIABLE_STYLESHEET.ts
```diff
@@ -0,0 +1,9 @@
+import { StylePropertyStyleSheet } from '../styles/StylePropertyStyleSheet'
+
+export const STYLE_VARIABLE_STYLESHEET = new StylePropertyStyleSheet(
+  (
+    document.querySelector('#initial-style-variables') as
```
erik-beus: Should we perhaps store this `id` somewhere in `packages/core` to ensure it's the same during SSR and client side setup? Also, consider if we should prefix it with `nordcraft`?

packages/runtime/src/components/createComponent.ts
```diff
@@ -68,6 +70,71 @@ export function createComponent({
...
+            selector,
+            styleVariable.name,
+          ),
+      },
+    )
+  })
+  node.variants?.forEach((variant) => {
+    variant['style-variables']?.forEach((styleVariable) => {
+      // style-variables on variants are always version 2
+      const signal = dataSignal.map((data) =>
```
erik-beus: It seems there's some potential for reusing code for this part and the part on line 75. The only difference is the selector for the variants it seems?
Reusing code might make it more reliable going forward when we introduce changes 🤞

packages/runtime/src/components/createElement.ts
```diff
@@ -130,21 +132,82 @@ export function createElement({
...
+        signal.subscribe((value) => elem.style.setProperty(`--${name}`, value))
+      }
+    }
+  })
+
+  node.variants?.forEach((variant) => {
+    variant['style-variables']?.forEach((styleVariable) => {
+      // style-variables on variants are always version 2
+      const { name, formula } = styleVariable
+      const signal = dataSignal.map((data) =>
```
erik-beus: This is also very similar to the code starting on line 139

packages/runtime/src/styles/StylePropertyStyleSheet.ts
```diff
@@ -0,0 +1,180 @@
...
+      this.styleSheet = styleSheet
+    } else {
+      this.styleSheet = new CSSStyleSheet()
+      document.adoptedStyleSheets.push(this.getStyleSheet())
+    }
+  }
+
+  /**
+   * @param selector The CSS selector to apply the property to.
+   * @param name The name of the property, without the `--` prefix.
```
erik-beus: Would it make sense to throw an error if the `name` did start with `--`?

packages/runtime/src/styles/StylePropertyStyleSheet.ts
```diff
@@ -0,0 +1,180 @@
...
+      let newRule = this.styleSheet.cssRules[ruleIndex]
+
+      // We are only interested in the dynamic style, so get the actual style rule, not media or other nested rules. Loop until we are at the bottom most rule.
+      while (
+        (newRule as any).cssRules &&
+        (newRule as CSSGroupingRule).cssRules.length > 0
+      ) {
+        newRule = (newRule as CSSGroupingRule).cssRules[0]
+      }
+      rule = newRule as CSSStyleRule | CSSNestedDeclarations
```
erik-beus: Is there any risk that this could be `null | undefined`?

packages/runtime/src/styles/StylePropertyStyleSheet.ts
```diff
@@ -0,0 +1,168 @@
...
+        rule = (rule as CSSGroupingRule).cssRules[0]
+      }
+
+      ruleIndex.set(selector, rule as CSSStyleRule)
+    }
+    return ruleIndex
+  }
+
+  private static selectorFromCSSRule(rule: CSSRule): string {
+    switch (rule.constructor.name) {
```
erik-beus: Would be nice if we had a better (builtin) type for this 🤔

packages/runtime/src/styles/StylePropertyStyleSheet.ts
```diff
@@ -0,0 +1,168 @@
...
+        console.warn(
+          `Unsupported CSS rule type: ${rule.constructor.name}. Returning empty selector.`,
+        )
+        return ''
+    }
+  }
+
+  private static getFullSelector(
+    selector: string,
+    options?: {
```
erik-beus: Do we also want to `console.warn` in this case?

packages/core/src/component/component.types.ts
```diff
@@ -75,6 +78,21 @@ export interface TextNodeModel {
...
 
+export type StyleVariable = {
+  version: 2
+  name: string
+  syntax: CssSyntaxNode
+  formula: Formula
+}
+
+export type StyleVariableLegacy = {
+  version: 1 | undefined
```
erik-beus: Would this be more accurate?

```suggestion
  version: never
```

packages/ssr/src/rendering/components.ts
```diff
@@ -397,7 +442,37 @@ const renderComponent = async ({
...
+        // Add extra instance styling for each style-variable
+        node['style-variables']?.forEach((styleVariable) => {
+          const value = applyFormula(styleVariable.formula, formulaContext)
+          if (isDefined(value)) {
+            addStyleVariable(
+              `[data-id="${path}"].${component.name}\\:${id}`,
+              `--${styleVariable.name}: ${applyFormula(
+                styleVariable.formula,
+                formulaContext,
+              )}`,
```
erik-beus:
```suggestion
              `--${styleVariable.name}: ${value}`,
```

packages/ssr/src/rendering/components.ts
```diff
@@ -397,7 +442,37 @@ const renderComponent = async ({
...
+            )
+          }
+        })
+
+        node.variants?.forEach((variant) => {
+          variant['style-variables']?.forEach((styleVariable) => {
+            const value = applyFormula(styleVariable.formula, formulaContext)
+            if (isDefined(value)) {
+              addStyleVariable(
+                `[data-id="${path}"].${component.name}\\:${id}${variantSelector(
```
erik-beus: It might be useful to have a shared utility function for generating these selectors across ssr/runtime

packages/ssr/src/rendering/components.ts
```diff
@@ -551,6 +641,33 @@ export const renderPageBody = async ({
   const apiCache: ApiCache = {}
   const updateApiCache = (key: string, value: ApiStatus) =>
     (apiCache[key] = value)
+  const styleVariables: Map<string, Set<string>> = new Map()
+  const addStyleVariable = (
+    selector: string,
+    rule: string,
```
erik-beus: Would it make sense to validate the pattern of this parameter? To ensure it starts with `--` for instance? Potentially with a template literal type https://www.typescriptlang.org/docs/handbook/2/template-literal-types.html#handbook-content

erik-beus (changes requested):
Please see suggestions

packages/core/src/styling/styleProperty.ts
jacob-kofoed: This link is more accurate for what we have https://developer.mozilla.org/en-US/docs/Web/CSS/@property/syntax

I think we are only going to support the primitives for now. The list at https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_Values_and_Units/CSS_data_types is mapping or combination of primitives.

packages/core/src/component/component.types.ts
jacob-kofoed: Done. This took a while, but an overall improvement in the editor especially.

packages/core/src/styling/styleProperty.ts
jacob-kofoed: I thought we were adding more besides primitives, but it will be a while if at all. Simplified it.

packages/runtime/src/components/STYLE_VARIABLE_STYLESHEET.ts
jacob-kofoed: Done ✅

packages/runtime/src/styles/StylePropertyStyleSheet.ts
jacob-kofoed: I changed the name to always include the double-dash. This should be handled with tooling in the editor, issue rules and one day a schema validator.

packages/runtime/src/components/createComponent.ts
jacob-kofoed: I combined the selector into its own function and the subscriber logic into its own one as well.

packages/runtime/src/styles/StylePropertyStyleSheet.ts
jacob-kofoed: Typescript doesn't complain when taking by index, which is usually a problem. In this case we get the index back as we insert, so the required type is fine.

packages/runtime/src/styles/StylePropertyStyleSheet.ts
jacob-kofoed: The built in type is deprecated. I couldn't find a better way after researching some more, but it is also a pretty obscure API, so perhaps it exists.

packages/ssr/src/rendering/components.ts
jacob-kofoed: I added a shared function. Should be much better now.

packages/core/src/styling/customProperty.ts
```diff
@@ -0,0 +1,74 @@
...
+  | 'url'
+
+export type CssSyntaxNode =
+  | {
+      type: 'primitive'
+      name: CssSyntax
+    }
+  | {
+      type: 'keyword'
+      keywords: string[]
```
erik-beus, line 28: Can keywords be any string?

packages/core/src/styling/style.css.ts
erik-beus: You resolved this one, but I can't find a test for this scenario?

packages/core/src/styling/theme.const.ts
```diff
@@ -1,5 +1,7 @@
 import type { OldTheme } from './theme'
 
+export const CUSTOM_PROPERTIES__STYLESHEET_ID = 'nc-custom-properties'
```
erik-beus:
```suggestion
export const CUSTOM_PROPERTIES_STYLESHEET_ID = 'nc-custom-properties'
```

packages/core/src/utils/getNodeSelector.ts
```diff
@@ -0,0 +1,28 @@
+import type { StyleVariant } from '../component/component.types'
+import { variantSelector } from '../styling/variantSelector'
+
+export function getNodeSelector(
```
erik-beus, line 16: It might be useful to add a few tests to this one. Mainly to explain what the final selectors will look like

packages/search/src/rules/style-variables/ambiguousStyleVariableSyntaxRule.ts
```diff
@@ -0,0 +1,151 @@
...
+    const { nodeType, value, files, memo, path } = args
+    if (nodeType !== 'component-node') {
+      return
+    }
+
+    if (value.type !== 'element' && value.type !== 'component') {
+      return
+    }
+
+    const allCustomPropertiesBySyntax = memo(
```
erik-beus: Could this potentially be a `Map` or `Set` to make lookups faster?

packages/core/src/utils/getNodeSelector.ts
jacob-kofoed: Done 👍

packages/search/src/rules/style-variables/ambiguousStyleVariableSyntaxRule.ts
jacob-kofoed: Don't think a Map is actually faster than a Record for lookups, but it makes it more obvious that we plan to use it as a lookup table. I changed it 👍

packages/core/src/styling/customProperty.ts
jacob-kofoed: Yes, for now the editor specifies what it can be.

packages/search/src/rules/style-variables/ambiguousStyleVariableSyntaxRule.ts
```diff
@@ -0,0 +1,155 @@
...
+    Object.entries(value.customProperties ?? {}).forEach(
+      ([customPropertyName, customProperty]) => {
+        const syntax = customProperty.syntax
+        if (syntax.type !== 'primitive') {
+          return
+        }
+
+        const duplicates = allCustomPropertiesBySyntax
+          .get(customPropertyName)
+          ?.filter(
```
erik-beus: My [comment](https://github.com/nordcraftengine/nordcraft/pull/477/files#r2197428492) was mainly to see if we could lookup duplicates in `O(1)` here?

jacob-kofoed: I'm storing first by name and then by syntax for faster lookups. I also changed it to only create the lookup table the first time an actual custom property is found.

I moved the memo function to a new folder as reuse of memoization key/function pairs might be useful across rules.

packages/search/src/rules/style-variables/ambiguousStyleVariableSyntaxRule.ts
```diff
@@ -0,0 +1,65 @@
...
+      if (syntax.type !== 'primitive') return
+      const allCustomPropertiesBySyntax = getAllCustomPropertiesBySyntax(memo, {
+        files,
+      })[propName]
+
+      const conflicts = Object.entries(allCustomPropertiesBySyntax)
+        .filter(([name]) => name !== syntax.name)
+        .flatMap(([, entries]) => entries)
+
+      if (conflicts.length) {
```
erik-beus: I suggest we stay explicit and avoid exploiting Javascript's faults ;)

```suggestion
      if (conflicts.length > 0) {
```

PR #493, opened 2025-07-31
Send uncaught preview rendering errors to the editor

jacob-kofoed:
This will let us show a toast when a user is using a circular formula with no exit condition or a recursive component. Unlike some other frameworks, I'm just using the browsers own maximum call stack error instead of a custom solution of keeping track of the render depth. Upside is less logic and a deeper max call stack. Downside is that we cannot explicitly say which component(s) or formula(s) are involved in the infinite depth. We could store a component stack in a separate global variable though 🤔 

This is implemented generically and can be used in the future for any info messages from the runtime to the editor.

To test, just add a component to itself or call a formula/action in themselves.

### DIFF
--- a/packages/runtime/src/debug/panicScreen.ts
+++ b/packages/runtime/src/debug/panicScreen.ts
@@ -1,0 +1,23 @@
+export const renderPanicScreen = (
+  title: string,
+  message: string,
+  isPage: boolean,
+) => `
+  <main style="background-color: blue; color: white; font-family: 'Courier New', monospace; font-weight: 100; display: flex; justify-content: flex-start; align-items: flex-start; margin: 0px; padding: 0px;">
+    <div style="display: flex; flex-direction: column; justify-content: space-between; align-items: flex-start; height: 100vh;">
+      <div style="display: flex; flex-direction: column; gap: 20px; padding: 80px;">
+        <div>
+          <h1 style="display: inline; font-size: 22px; background: white; color: blue; padding: 8px 16px;">${title}</h1>
+        </div>
+        <div>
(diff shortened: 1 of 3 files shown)


packages/runtime/src/editor-preview.main.ts
```diff
@@ -1406,19 +1407,39 @@ export const createRoot = (
...
+            parentElement: domNode,
+            instance: { [newCtx.component.name]: 'root' },
+          })
+          newCtx.component.onLoad?.actions.forEach((action) => {
+            // eslint-disable-next-line @typescript-eslint/no-floating-promises
+            handleAction(action, dataSignal.get(), newCtx)
+          })
+          rootElem.forEach((elem) => domNode.appendChild(elem))
+        } catch (error: unknown) {
+          let title = 'Unexpected error while rendering component'
```
erik-beus: Perhaps we could differentiate between pages and components in the title?

erik-beus (changes requested):
Looks good. Small suggestion. I'll test now 👍

jacob-kofoed:
image

I added a version of our blue screen of death to unrecoverable errors, on your request :)

erik-beus:
> image
> 
> 
> 
> I added a version of our blue screen of death to unrecoverable errors, on your request :)

I love it 🙌

packages/runtime/src/debug/panicScreen.ts
```diff
@@ -0,0 +1,37 @@
...
+  name: string
+  message: string
+  isPage?: boolean
+  cause?: unknown
+}) => {
+  let content = getContent(name, message, isPage)
+
+  // Easter egg for RangeError
+  if (cause instanceof RangeError) {
+    for (let i = 0; i < 10; i++) {
```
jacob-kofoed, line 16: Kasper signed off on this. Just for the RangeError:

image

PR #494, opened 2025-07-31
Pseudo element preview

jacob-kofoed:
- Fixed an issue where pseudo elements would not preview correctly in the editor
- Web components works with custom style properties v2

### DIFF
--- a/packages/runtime/src/utils/subscribeCustomProperty.ts
+++ b/packages/runtime/src/utils/subscribeCustomProperty.ts
@@ -22,16 +22,14 @@ export function subscribeCustomProperty({
   root: Document | ShadowRoot
   runtime: Runtime
 }) {
-  customPropertiesStylesheet =
-    customPropertiesStylesheet ??
-    new CustomPropertyStyleSheet(
-      root,
-      (
-        root.getElementById(CUSTOM_PROPERTIES_STYLESHEET_ID) as
-          | HTMLStyleElement
-          | undefined
-      )?.sheet,
-    )
+  customPropertiesStylesheet ??= new CustomPropertyStyleSheet(
+    root,
+    (
+      root.getElementById(CUSTOM_PROPERTIES_STYLESHEET_ID) as
+        | HTMLStyleElement
+        | undefined
+    )?.sheet,
+  )
 
   signal.subscribe(
     customPropertiesStylesheet.registerProperty(


packages/runtime/src/styles/CustomPropertyStyleSheet.ts
```diff
@@ -44,7 +44,10 @@ export class CustomPropertyStyleSheet {
     // Check if the selector already exists
     let rule = this.ruleMap.get(fullSelector)
     if (!rule) {
-      const ruleIndex = this.styleSheet.insertRule(fullSelector)
+      const ruleIndex = this.styleSheet.insertRule(
+        fullSelector,
+        this.styleSheet.cssRules.length,
```
erik-beus, line 49: Was this to ensure that new rules are added last?

packages/runtime/src/utils/subscribeCustomProperty.ts
```diff
@@ -4,38 +4,52 @@ import type { Signal } from '../signal/signal'
...
   variant,
+  root,
+  runtime,
 }: {
   selector: string
   customPropertyName: string
   signal: Signal<string>
   variant?: StyleVariant
+  root: Document | ShadowRoot
+  runtime: 'page' | 'custom-element' | 'preview'
```
erik-beus: Could we reuse the runtime type from somewhere else perhaps? Just to ensure we only declare it once

packages/runtime/src/utils/subscribeCustomProperty.ts
```diff
@@ -4,38 +4,52 @@ import type { Signal } from '../signal/signal'
...
 }) {
+  customPropertiesStylesheet =
+    customPropertiesStylesheet ??
```
erik-beus:
```suggestion
  customPropertiesStylesheet ??=
```

erik-beus (approved):
Looks good. Just a few questions/suggestions

packages/runtime/src/utils/subscribeCustomProperty.ts
jacob-kofoed: Good suggestion. I added a shared type to core.

packages/runtime/src/styles/CustomPropertyStyleSheet.ts
jacob-kofoed: I was fairly certain I had it correct before, but the order of variants was opposite, so this should correct it. It is still guarded by feature-flag, so there's room to fix things still. I will try to add some tests for this as well, but the mock implementation in our testing-framework does not reflect the real spec very well...

PR #500, opened 2025-08-11
Add issue rule for invalid interactive content

erik-beus:
Add a warning when a button or `<a>` element has interactive content as a child element. This is not valid and should be avoided. There are edge cases when using slots and conditionals that might lead to more warnings than necessary, but generally this rule should help improve correctness of users' application structure.

### DIFF
--- a/packages/search/src/rules/dom/elementWithoutInteractiveContentRule.ts
+++ b/packages/search/src/rules/dom/elementWithoutInteractiveContentRule.ts
@@ -1,7 +1,7 @@
 import type { Component } from '@nordcraft/core/dist/component/component.types'
 import type { ToddleComponent } from '@nordcraft/core/dist/component/ToddleComponent'
 import { isDefined } from '@nordcraft/core/dist/utils/util'
-import type { Level, Rule } from '../../types'
+import type { Rule } from '../../types'
 import {
   interactiveContentElementDefinition,
   type InteractiveContent,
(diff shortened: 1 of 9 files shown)


packages/search/src/rules/dom/elementWithoutInteractiveContentRule.ts
```diff
@@ -0,0 +1,87 @@
...
+import type { Level, Rule } from '../../types'
+import {
+  interactiveContentElementDefinition,
+  type InteractiveContent,
+} from '../../util/helpers'
+
+const ELEMENTS_WITHOUT_INTERACTIVE_CONTENT = ['button', 'a']
+
+export function elementWithoutInteractiveContentRule({
+  level = 'warning',
```
jacob-kofoed: Not sure why you need to set this as an attribute? And isn't this problem potentially app-breaking, so should be at error-level?

packages/search/src/util/helpers.ts
```diff
@@ -43,3 +47,61 @@ const LEGACY_CUSTOM_ACTIONS = new Set([
...
+  INTERACTIVE_CONTENT.find((ic) => {
+    if (element.tag !== ic.tag) {
+      return false
+    }
+    if ('whenAttributeIsPresent' in ic) {
+      return isDefined(element.attrs[ic.whenAttributeIsPresent])
+    }
+    if ('whenAttributeIsNot' in ic) {
+      const attributeFormula = element.attrs[ic.whenAttributeIsNot.attribute]
+      if (
```
jacob-kofoed: You could potentially simplify a bit by skipping the if here.

```ts
return attributeFormula && attributeFormula.type === 'value' && attributeFormula.value === ic.whenAttributeIsNot.value;
```

PR #504, opened 2025-08-13
fix: specify isFormula type

mikkelnordcraft:
This will specify the isFormula check to check for actual formula types.

### DIFF
--- a/packages/core/src/formula/formula.ts
+++ b/packages/core/src/formula/formula.ts
@@ -142,7 +142,23 @@ export type ToddleEnv =
     }
 
 export function isFormula(f: any): f is Formula {
-  return f && typeof f === 'object' && typeof f.type === 'string'
+  return (
+    f &&
+    typeof f === 'object' &&
+    typeof f.type === 'string' &&
+    [
+      'path',
+      'function',
+      'record',
+      'object',
+      'array',
+      'or',
+      'and',
+      'apply',
+      'value',
+      'switch',
+    ].includes(f.type)
+  )
 }
 export function isFormulaApplyOperation(
   formula: Formula,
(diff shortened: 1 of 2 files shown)


packages/core/src/formula/formula.ts
```diff
@@ -142,7 +142,23 @@ export type ToddleEnv =
     }
 
 export function isFormula(f: any): f is Formula {
```
erik-beus, line 144: This might be slightly more accurate
```suggestion
export function isFormula(f: unknown): f is Formula {
```

packages/core/src/formula/formula.ts
```diff
@@ -142,7 +142,23 @@ export type ToddleEnv =
...
+      'function',
+      'record',
+      'object',
+      'array',
+      'or',
+      'and',
+      'apply',
+      'value',
+      'switch',
+    ].includes(f.type)
```
erik-beus: Just to make sure we don't include something that's not a valid type for a formula, consider this type casting
```suggestion
    ([
      'path',
      'function',
      'record',
      'object',
      'array',
      'or',
      'and',
      'apply',
      'value',
      'switch',
    ] as Formula['type'][]).includes(f.type)
```

erik-beus (changes requested):
Please see suggestions

packages/core/src/formula/formula.ts
mikkelnordcraft: Typescript does not like unknown. It refuses the type property.

packages/core/src/formula/formula.ts
mikkelnordcraft: Yes! Great idea.

erik-beus (approved):

PR #505, opened 2025-08-13
GraphQL introspection support in editor/preview

erik-beus:
- Add support for iframe event: "introspect_qraphql_api"
- Return introspection result to the parent window with the introspection result and the API key

### DIFF
--- a/packages/runtime/src/editor-preview.main.ts
+++ b/packages/runtime/src/editor-preview.main.ts
@@ -3,6 +3,7 @@
 /* eslint-disable no-case-declarations */
 /* eslint-disable no-fallthrough */
 import { isLegacyApi } from '@nordcraft/core/dist/api/api'
+import { type ApiRequest } from '@nordcraft/core/dist/api/apiTypes'
 import type {
   AnimationKeyframe,
   Component,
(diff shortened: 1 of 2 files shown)


packages/runtime/src/editor-preview.main.ts
```diff
@@ -1796,3 +1771,91 @@ const insertTheme = (parent: HTMLElement, theme: Theme | OldTheme) => {
   })
   parent.appendChild(styleElem)
 }
+
+type PostMessageType =
```
jacob-kofoed, line 1775: Nice with types here. Perhapsihaps each message type would be a good candidate for its own file, as each message has very little to do with each other. I'll add it to my todo. Would be nice before this file reaches 2k lines 😱

packages/runtime/src/editor/graphql.ts
```diff
@@ -0,0 +1,167 @@
...
+fragment InputValue on __InputValue {
+  name
+  description
+  type { ...TypeRef }
+  defaultValue
+}
+fragment TypeRef on __Type {
+  kind
+  name
+  ofType {
```
jacob-kofoed, line 69: GraphQL recursion is beautiful

PR #507, opened 2025-08-15
Add group property to all actions

whitep4nth3r:
Add group property to all actions to enable grouping in the UI.

1st stage of https://github.com/orgs/nordcraftengine/projects/19/views/1?pane=issue&itemId=123494939&issue=nordcraftengine%7Cnordcraft-internal%7C2020

### DIFF
--- a/package.json
+++ b/package.json
@@ -2,7 +2,7 @@
   "name": "nordcraft",
   "type": "module",
   "license": "Apache-2.0",
-  "version": "1.0.35",
+  "version": "1.0.36",
   "homepage": "https://github.com/nordcraftengine/nordcraft",
   "private": "false",
   "workspaces": [


erik-beus (approved):

packages/lib/schemas/group.schema.json
```diff
@@ -0,0 +1,23 @@
+{
+  "$id": "group.schema.json",
```
erik-beus, line 2: Consider calling it something like this instead?

```suggestion
  "$id": "libAction.schema.json",
```

packages/lib/schemas/group.schema.json
```diff
@@ -0,0 +1,23 @@
...
+  "$schema": "https://json-schema.org/draft/2020-12/schema",
+  "title": "Group Property",
+  "description": "The group to which an action belongs. This is used to categorize actions in the front end UI.",
```
erik-beus, line 5:
```suggestion
  "description": "A lib action used in the Nordcraft runtime to perform specific operations.",
```

PR #509, opened 2025-08-15
Update capitalization of action names and add new sharing category

whitep4nth3r:


packages/lib/actions/clearLocalStorage/action.json
```diff
@@ -1,6 +1,6 @@
 {
   "$schema": "../../schemas/libAction.schema.json",
-  "name": "Clear Local Storage",
+  "name": "Clear local storage",
```
erik-beus, line 3: I'm quite sure this will break all references to this action. I believe we introduced a display name or label in formulas. Perhaps we can do something similar for actions?

erik-beus (changes requested):

packages/lib/actions/clearLocalStorage/action.json
whitep4nth3r: I believe it only broke when we changed the actual file names before?

If this is the case though, I'd rather just fix it with CSS, but it looks like there may be a bug in the CSS parser as ::first-letter wasn't working when I tried it earlier today.

erik-beus: I just tested locally, and the renaming doesn't affect the actions' functionality 👍 They're keyed by the folder name

whitep4nth3r: p.s. no bug in the CSS parser, it's the way the ActionSelect is set up with an image before the text and so first-letter can't target the text.

PR #515, opened 2025-08-22
Add ability to fix issue rules on worker thread

erik-beus:
- Add fixes for `legacyFormulaRule`
- Add fix for `noReferenceComponentRule`
- Add a few tests
- Add function overloads for `searchProject` to allow reusing code for applying fixes
- Adjust `problems.worker.ts` to support fix messages

### DIFF
--- a/packages/search/src/fixProject.ts
+++ b/packages/search/src/fixProject.ts
@@ -1,0 +1,45 @@
+import type { ProjectFiles } from '@nordcraft/ssr/dist/ssr.types'
+import { searchProject } from './searchProject'
+import type { ApplicationState, FixType, Rule } from './types'
+
+/**
+ * Uses searchProject to apply 1 fix at a time. Finally ends up with the resulting
+ * ProjectFiles that are returned.
+ * @returns ProjectFiles - updated files after fixes have been applied (if any)
+ */
+export const fixProject = ({
+  files,
+  rule,
+  fixType,
+  pathsToVisit = [],
+  state,
+}: {
+  files: Omit<ProjectFiles, 'config'> & Partial<Pick<ProjectFiles, 'config'>>
+  rule: Rule
+  fixType: FixType
+  pathsToVisit?: string[][]
+  state?: ApplicationState
+}) => {
+  let updatedFiles = files
+  let iterations = 0
+  const maxIterations = 100
+  // Apply 1 fix per iteration for ${maxIterations} iterations
+  // After every fix, the updated files are used for the next iteration
+  while (iterations < maxIterations) {
+    iterations++
+    const suggestedFiles = searchProject({
+      files: updatedFiles,
+      rules: [rule],
+      options: { mode: 'FIX', fixType },
+      pathsToVisit,
+      state,
+    }).next()
+    if (suggestedFiles.value) {
+      updatedFiles = suggestedFiles.value
+    } else {
+      // No more fixes could be applied
+      break
+    }
+  }
+  return updatedFiles
+}
--- a/packages/search/src/problems.worker.ts
+++ b/packages/search/src/problems.worker.ts
@@ -1,4 +1,6 @@
 import type { ProjectFiles } from '@nordcraft/ssr/dist/ssr.types'
+import { compare } from 'fast-json-patch'
+import { fixProject } from './fixProject'
 import { createActionNameRule } from './rules/actions/createActionNameRule'
 import { legacyActionRule } from './rules/actions/legacyActionRule'
 import { noReferenceProjectActionRule } from './rules/actions/noReferenceProjectActionRule'
@@ -56,7 +58,15 @@ import { unknownTriggerWorkflowParameterRule } from './rules/workflows/unknownTr
 import { unknownTriggerWorkflowRule } from './rules/workflows/unknownTriggerWorkflowRule'
 import { unknownWorkflowParameterRule } from './rules/workflows/unknownWorkflowParameterRule'
 import { searchProject } from './searchProject'
-import type { ApplicationState, Category, Code, Level, Result } from './types'
+import type {
+  ApplicationState,
+  Category,
+  Code,
+  FixType,
+  Level,
+  Result,
+  Rule,
+} from './types'
 
 export type Options = {
   /**
@@ -177,13 +187,23 @@ const RULES = [
   ambiguousStyleVariableSyntaxRule,
 ]
 
-/**
- * This function is a web worker that checks for problems in the files.
- */
-onmessage = (
-  event: MessageEvent<{ files: ProjectFiles; options?: Options }>,
-) => {
-  const { files, options = {} } = event.data
+interface FindProblemsArgs {
+  files: ProjectFiles
+  options?: Options
+}
+
+interface FixProblemsArgs {
+  files: ProjectFiles
+  options?: Options
+  fixRule: Rule
+  fixType: FixType
+  id: string
+}
+
+type Message = FindProblemsArgs | FixProblemsArgs
+
+const findProblems = (data: FindProblemsArgs) => {
+  const { files, options = {} } = data
   const rules = RULES.filter(
     (rule) =>
       (!options.categories || options.categories.includes(rule.category)) &&
@@ -232,4 +252,36 @@ onmessage = (
 
   // Send the remaining results
   postMessage(batch)
+}
+
+const fixProblems = (data: FixProblemsArgs) => {
+  const { files, options = {} } = data
+
+  const updatedFiles = fixProject({
+    files,
+    rule: data.fixRule,
+    fixType: data.fixType,
+    pathsToVisit: options.pathsToVisit,
+    state: options.state,
+  })
+  // Calculate diff
+  const diff = compare(files, updatedFiles)
+  // Send diff + metadata to main thread
+  postMessage({
+    id: data.id,
+    patch: diff,
+    rule: data.fixRule,
+    fixType: data.fixType,
+  })
+}
+
+/**
+ * This function is a web worker that checks for problems in the files.
+ */
+onmessage = (event: MessageEvent<Message>) => {
+  if ('fixRule' in event.data) {
+    fixProblems(event.data)
+  } else {
+    findProblems(event.data)
+  }
 }
--- a/packages/search/src/rules/components/noReferenceComponentRule.ts
+++ b/packages/search/src/rules/components/noReferenceComponentRule.ts
@@ -1,39 +1,55 @@
 import type { Component } from '@nordcraft/core/dist/component/component.types'
+import { omit } from '@nordcraft/core/dist/utils/collections'
 import { isDefined } from '@nordcraft/core/dist/utils/util'
-import type { Rule } from '../../types'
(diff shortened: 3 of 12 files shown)


packages/search/src/searchProject.ts
```diff
@@ -53,13 +75,14 @@ export function* searchProject({
           memo,
         },
         state,
-      )
+        options: options as any,
```
erik-beus: I tried in various ways to make Typescript happy, but this seemed like the easiest fix. Even though I use function overloading, it would complain about union types when I called `visitNode`.

packages/core/src/formula/formulaTypes.ts
```diff
@@ -31,7 +32,9 @@ export type PluginFormula<Handler = string | Function> =
 
 export const isToddleFormula = <Handler>(
   formula: PluginFormula<Handler>,
-): formula is ToddleFormula => Object.hasOwn(formula, 'formula')
+): formula is ToddleFormula =>
+  Object.hasOwn(formula, 'formula') &&
+  isDefined((formula as ToddleFormula).formula)
```
jacob-kofoed, line 37: What was the issue here? Isn't it kind of doing the same thing?

packages/search/src/fixProject.ts
```diff
@@ -0,0 +1,48 @@
...
+  pathsToVisit?: string[][]
+  useExactPaths?: boolean
+  state?: ApplicationState
+}) => {
+  let updatedFiles = files
+  let iterations = 0
+  const maxIterations = 100
+  // Apply 1 fix per iteration for ${maxIterations} iterations
+  // After every fix, the updated files are used for the next iteration
+  while (iterations < maxIterations) {
```
jacob-kofoed:
```ts
while(iterations++ < maxIterations) {
   ...
}
```

packages/search/src/fixProject.ts
```diff
@@ -0,0 +1,48 @@
...
+    iterations++
+    const suggestedFiles = searchProject({
+      files: updatedFiles,
+      rules: [rule],
+      options: { mode: 'FIX', fixType },
+      pathsToVisit,
+      useExactPaths,
+      state,
+    }).next()
+    if (suggestedFiles.value) {
```
jacob-kofoed: I suppose `suggestedFiles.done` could not be used here, as you use undefined as a valid return value?

You could also consider destructing the `suggestedFiles` into `const { suggestedFiles: value } = searchProject` as the name doesn't really indicate that the const is a generator instance.

packages/search/src/problems.worker.ts
```diff
@@ -231,5 +271,45 @@ onmessage = (
   }
 
   // Send the remaining results
-  postMessage(batch)
+  respond(batch)
+}
+
+const fixProblems = (data: FixProblemsArgs) => {
+  const { files, options = {} } = data
```
jacob-kofoed: Why only deconstruct parts of data? You could deconstuct `fixRule` etc. as well here

packages/search/src/problems.worker.ts
```diff
@@ -231,5 +271,45 @@ onmessage = (
...
+  const rule = RULES.find((r) => r.code === data.fixRule)
+  if (!rule) {
+    // eslint-disable-next-line no-console
```
jacob-kofoed: I think throwing is fine here. We could potentially catch exceptions at the top-level and post back to the editor to show that something went wrong in a toast or something eventually.

packages/search/src/problems.worker.ts
```diff
@@ -231,5 +271,45 @@ onmessage = (
...
+    rule,
+    fixType: data.fixType,
+    pathsToVisit: options.pathsToVisit,
+    useExactPaths: options.useExactPaths,
+    state: options.state,
+  })
+  // Calculate diff
+  const jsonDiffPatch = create({ omitRemovedValues: true })
+  const diff = jsonDiffPatch.diff(files, updatedFiles)
+  // Send diff + metadata to main thread
```
jacob-kofoed, line 295: Does it make sense to check for an empty diff here? Or is it handled later at socket-level etc? To avoid pushing changes to the history?

Can we actually undo fixes, or do they skip the client-history even when on the active component?

packages/search/src/rules/components/noReferenceComponentRule.ts
```diff
@@ -1,40 +1,56 @@
...
-      isPage(value) ||
-      (state?.projectDetails?.type === 'package' && value.exported === true)
-    ) {
+  visit: (report, data, state) => {
+    if (isValid(data, state)) {
       return
     }
+    report(data.path, undefined, ['delete-component'])
+  },
+  fix: (data, fixType, state) => {
```
jacob-kofoed: Would it make sense to make this more data-driven, instead of having to manually perform checks etc.

We could automatically call `visit` before a fix and make sure that the visit still reports the issue with the correct code and that the fix-type is in the array of available fixes.

Then fix could be an object with keys matching the fix-types. Ie.

```ts
fix: {
  "delete-component": (data, state) => { ... },
  "other-fix": (data, state) => { ... }
```

Either approach will work, but we are going to write a lot of fixes so having a strict structure may save us problems in the future.

packages/search/src/rules/components/noReferenceComponentRule.ts
```diff
@@ -1,40 +1,56 @@
...
-        if (
-          node.type === 'component' &&
-          node.name === value.name &&
-          // Circular references from a component to itself should
-          // not count as a reference
-          node.name !== component!.name
-        ) {
-          return
-        }
+const isValid = (data: NodeType, state?: ApplicationState) => {
```
jacob-kofoed: `isValid` is a bit confusing naming to me either way. Is it a valid issue or a valid component?

I suggest something like "hasIssue" or specific for this type "isReferenced"

packages/search/src/searchProject.ts
```diff
@@ -42,151 +68,213 @@ export function* searchProject({
...
+  state: ApplicationState | undefined
+  options?: { mode: 'FIX'; fixType: FixType }
+}): Generator<Result | ProjectFiles | void> {
   performance.mark(`visitNode-${args.path.join('/')}`)
-  const { rules, pathsToVisit, ...data } = args
+  const { rules, pathsToVisit, useExactPaths, ...data } = args
   const { files, value, path, memo, nodeType } = data
-  if (!shouldSearchPath(data.path, pathsToVisit)) {
+  if (
+    !shouldSearchPath({
```
jacob-kofoed: Any way to combine `shouldSearchPath` and `shouldSearchExactPath` into one function that takes the `useExactPaths` as an arg? This logic is a little confusing for me to understand at a glance, as we first check one and then the other.

packages/search/src/searchProject.ts
```diff
@@ -42,151 +68,213 @@ export function* searchProject({
...
+      for (const rule of rules) {
+        const fixedFiles = rule.fix?.(data, options.fixType, state)
+        if (fixedFiles) {
+          yield fixedFiles
+        }
+      }
+    } else {
+      const results: Result[] = []
+      for (const rule of rules) {
+        performance.mark(`rule-${rule.code}`)
```
jacob-kofoed: Just a suggestion if you find it interesting: You can do `performance.measure` if you want a timeline for this or create your own timing track using the performance API or console.timeStamp with `performance.now()` before/after. I think console.timeStamp has much less overhead as it does not push performance entries when not recording via. devtools. See: https://developer.chrome.com/docs/devtools/performance/extension

packages/search/src/searchProject.ts
```diff
@@ -42,151 +68,213 @@ export function* searchProject({
...
-      },
-      data,
-      state,
-    )
-  }
+  if (
+    !useExactPaths ||
+    shouldSearchExactPath({ path: data.path, pathsToVisit })
+  ) {
+    if (options) {
```
jacob-kofoed: This check to see if we are to fix or search is a little hidden. Maybe a helper method `shouldFix()` or call `options` `fixOptions` instead.

packages/search/src/searchProject.ts
```diff
@@ -395,23 +505,24 @@ function* visitNode(
...
           formula: f,
         } of formula.formulasInFormula()) {
-          yield* visitNode(
-            {
+          yield* visitNode({
+            args: {
               nodeType: 'formula',
               value: f,
-              path: [...path, ...formulaPath],
+              path: [...path, 'formula', ...formulaPath],
```
jacob-kofoed, line 518: This seems like a quite serious issue we had? Surprised we only noticed this now? Does deep-linking from issues to formulas still work, or maybe it didn't work?

jacob-kofoed (changes requested):

packages/core/src/formula/formulaTypes.ts
erik-beus: In my (old) project, I had some global formulas that had a `formula` field defined, but the value was `null`. I'm not sure when this happens, but I think it's good to make our type guard as safe as possible.

packages/search/src/problems.worker.ts
erik-beus: In the editor I check for whether the diff is defined. The variable in the editor that holds information about the "active" fix will be reset regardless when the message/response is received from the fix.
Currently, I do no distinguish between changes to the active component and to other files in the editor, so the undo history won't work atm. Perhaps we can fix that before we enable the autofix feature for everyone though.

packages/search/src/searchProject.ts
erik-beus: They do 2 quite separate things, so we would probably need to have it return an object with 2 different values. I've renamed the first one instead, to make it more clear that it tells us whether we should visit a tree/folder.

packages/search/src/rules/components/noReferenceComponentRule.ts
erik-beus: Adjusted

packages/search/src/rules/components/noReferenceComponentRule.ts
erik-beus: I've adjusted the structure.
Automatically calling visit might work, but with the current type setup, it's actually quite useful to have a "validation" function before the fix to act as a type guard. That makes it easier to work with the data in a typesafe way in the actual fix implementation.
Let me know if you think it's alright as it is now.

packages/search/src/fixProject.ts
erik-beus: Yes exactly. I've destructured it now.

packages/search/src/searchProject.ts
erik-beus: I don't think it ever worked, and I'm not sure if we have even been visiting nodes inside of global formulas. This would mean that legacy formulas used in global formulas, for instance, would never have been reported.

packages/search/src/rules/components/noReferenceComponentRule.ts
jacob-kofoed: I get the type guard problem. Logically we are going to check the same thing in the visit -> report as we are in the fix in all cases I can think of. Lets do it like this for now. As long as we keep writing test-cases for fixes then I think we are safe from any mistakes that a more data-driven approach would help catch as well.

PR #521, opened 2025-08-28
Fix/global formulas crud with open canvas

ChrEsb:
This PR introduces a new message type in the preview environment. With the introduction of services we can create global formulas without leaving the iframe. That meant that new global formulas were never passed to the iframe as they were only loaded on load. So to fix this we introduce a new message that will update the global formulas when notified.


packages/runtime/src/editor-preview.main.ts
```diff
@@ -66,6 +66,11 @@ type ToddlePreviewEvent =
       type: 'style_variant_changed'
       variantIndex: number | null
     }
+  | {
+      type: 'global_formulas'
+      formulas: Record<string, PluginFormula<FormulaHandlerV2>>
+      packageName: string
```
erik-beus: Is package name relevant when we can only modify formulas within the same project?

packages/runtime/src/editor-preview.main.ts
```diff
@@ -66,6 +66,11 @@ type ToddlePreviewEvent =
       type: 'style_variant_changed'
       variantIndex: number | null
     }
+  | {
+      type: 'global_formulas'
```
erik-beus: Should we also support global actions? For instance if you delete a global action from the left menu, that wouldn't necessarily mean you leave the iframe?

erik-beus (changes requested):

packages/runtime/src/editor-preview.main.ts
```diff
@@ -415,6 +438,43 @@ export const createRoot = (
...
+          const formulas: Record<string, PluginFormula<FormulaHandlerV2>> = {}
+          window.toddle.wipeLegacyFormulas?.()
+          Object.entries(message.data.formulas ?? {}).forEach(
+            ([name, formula]) => {
+              if (
+                !isToddleFormula<FormulaHandlerV2 | string>(formula) &&
+                typeof formula.name === 'string' &&
+                formula.version === undefined
+              ) {
+                Function(formula.handler as unknown as string)()
```
erik-beus, line 452: A comment here might be useful to explain that they're self-registering

packages/runtime/src/editor-preview.main.ts
```diff
@@ -175,21 +184,28 @@ export const initGlobalObject = ({
...
+      formulas: {},
+      actions: {},
       registerAction: (name, handler) => {
         if (legacyActions[name]) {
           console.error('There already exists an action with the name ', name)
           return
         }
         legacyActions[name] = handler
       },
+      wipeLegacyActions: () => {
```
erik-beus: `clear` might be more descriptive
```suggestion
      clearLegacyActions: () => {
```

packages/runtime/src/editor-preview.main.ts
```diff
@@ -201,6 +217,13 @@ export const initGlobalObject = ({
           argumentInputDataList[name] = getArgumentInputData
         }
       },
+      wipeLegacyFormulas: () => {
```
erik-beus:
```suggestion
      clearLegacyFormulas: () => {
```

bun.lock
```diff
@@ -101,7 +101,7 @@
 
     "@dmsnell/diff-match-patch": ["@dmsnell/diff-match-patch@1.1.0", "", {}, "sha512-yejLPmM5pjsGvxS9gXablUSbInW7H976c/FJ4iQxWIm7/38xBySRemTPDe34lhg1gVLbJntX0+sH0jYfU+PN9A=="],
 
-    "@emnapi/runtime": ["@emnapi/runtime@1.4.4", "", { "dependencies": { "tslib": "^2.4.0" } }, "sha512-hHyapA4A3gPaDCNfiqyZUStTMqIkKRshqPIuDOXv1hcBnD4U3l8cP0T1HMCfGRxQ6V64TGCcoswChANyOAwbQg=="],
+    "@emnapi/runtime": ["@emnapi/runtime@1.5.0", "", { "dependencies": { "tslib": "^2.4.0" } }, "sha512-97/BJ3iXHww3djw6hYIfErCZFee7qCtrneuLa20UXFCOTCfBM2cvQHjWJ2EG0s0MtdNwInarqCTz35i4wWXHsQ=="],
```
erik-beus: I suppose you could omit your changes to `bun.lock` as you have not changed any dependencies?

PR #525, opened 2025-09-02
Stop supporting autofix of BOOLEAN formula

erik-beus:
The implementation of the `BOOLEAN` formula was `Boolean(value)` which doesn't match the new formula's implementation (`value !== false && value !== undefined && value !== null`). Therefore, we cannot autofix it safely.


jacob-kofoed:
Good to know 👍 I wouldn't have guessed BOOLEAN formula to differ 🤷

PR #526, opened 2025-09-02
Allow autofixing legacy actions

erik-beus:
Add support for autofixing uses of these legacy actions:

- If
- PreventDefault
- StopPropagation
- UpdateVariable
- SetTimeout
- SetInterval
- Debug
- GoToURL
- TriggerEvent

I did not add support for fixing these legacy actions as they require manual intervention by the user:

- CopyToClipboard
- Update URL parameter
- Fetch
- FocusElement
- UpdateVariable
- TriggerEvent
- @toddle/setSessionCookies

I also updated our types to better reflect the data we see in actual projects.

### DIFF
--- a/packages/search/src/rules/actions/legacyActionRule.test.ts
+++ b/packages/search/src/rules/actions/legacyActionRule.test.ts
@@ -1,8 +1,14 @@
+import type {
+  ActionModel,
+  ElementNodeModel,
+} from '@nordcraft/core/dist/component/component.types'
+import type { ProjectFiles } from '@nordcraft/ssr/dist/ssr.types'
 import { describe, expect, test } from 'bun:test'
+import { fixProject } from '../../fixProject'
 import { searchProject } from '../../searchProject'
 import { legacyActionRule } from './legacyActionRule'
 
-describe('legacyAction', () => {
+describe('find legacyActions', () => {
   test('should detect legacy actions used in components', () => {
     const problems = Array.from(
       searchProject({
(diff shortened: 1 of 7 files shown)


packages/search/src/rules/actions/legacyActionRule.test.ts
```diff
@@ -90,3 +96,178 @@ describe('legacyAction', () => {
...
+    }
+    const fixedProject = fixProject({
+      files: projectFiles,
+      rule: legacyActionRule,
+      fixType: 'replace-legacy-action',
+    })
+    const fixedAction = (
+      fixedProject.components['apiComponent']?.nodes['root'] as ElementNodeModel
+    ).events['click'].actions[0]
+    expect(fixedAction).toBeDefined()
```
jacob-kofoed: Me being lazy would probably just use "toMatchInlineSnapshot()" instead of all this hard manual labor.

packages/search/src/util/helpers.ts
```diff
@@ -58,22 +63,19 @@ const LEGACY_CUSTOM_ACTIONS = new Set([
...
-  'Update Query',
   'Fetch',
   'SetTimeout',
   'SetInterval',
   'FocusElement',
   'Debug',
   'GoToURL',
   'TriggerEvent',
-  'Set session cookies',
+  '@toddle/setSessionCookies',
```
jacob-kofoed, line 79: Is this correct? We only have exactly one legacy action prefixed with `@toddle/`?

jacob-kofoed (approved):

packages/search/src/util/helpers.ts
erik-beus: That is correct. When we added the new `@toddle/setCookie` action, I thought I deprecated the old `Set session cookies` action. But I forgot that the previous one was not a legacy action - it was in fact built into our lib.
When we have proper runtime versioning, we should try to get rid of the old `@toddle/setSessionCookies` action 🤞

packages/search/src/rules/actions/legacyActionRule.test.ts
erik-beus: It was really hard 😂 But yes, it's probably easier to read it using inline snapshots 👍

PR #531, opened 2025-09-03
Make sure we load custom actions and formulas correctly in preview

ChrEsb:


packages/ssr/src/custom-code/codeRefs.ts
```diff
@@ -262,7 +262,7 @@ export const formulas = {
  * @param name
  * @returns "safe" function name only containing alphanumeric characters and _, e.g. "myFunction" or "my_function"
  */
-const safeFunctionName = (name: string) => {
+export const safeFunctionName = (name: string) => {
```
erik-beus: I suggest moving this to `core` if you want to use it in the `runtime` package. Mainly to avoid bloating the runtime's bundle size

erik-beus (changes requested):
Just 1 suggestion

erik-beus (approved):

PR #534, opened 2025-09-03
Add MDN documentation links to formulas

codenitrr:
Added MDN links to most of the formulas.

### DIFF
--- a/packages/lib/formulas/add/formula.json
+++ b/packages/lib/formulas/add/formula.json
@@ -17,6 +17,7 @@
   ],
   "variableArguments": true,
   "description": "Get the sum of multiple numbers.",
+  "mdnUrl": "https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Add",
   "output": {
     "description": "The sum of all the input numbers.",
     "type": { "type": "Number" }
--- a/packages/lib/formulas/add/formula.json
+++ b/packages/lib/formulas/add/formula.json
@@ -27,7 +27,7 @@
   ],
   "variableArguments": true,
   "description": "Get the sum of multiple numbers.",
-  "mdnUrl": "https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Add",
+  "mdnUrl": "https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Addition",
   "output": {
     "description": "The sum of all the input numbers.",
     "type": {
--- a/packages/lib/formulas/add/formula.json
+++ b/packages/lib/formulas/add/formula.json
@@ -5,14 +5,24 @@
     {
       "name": "0",
       "description": "Number to be added.",
-      "type": { "type": "Number" },
-      "formula": { "type": "value", "value": 1 }
+      "type": {
+        "type": "Number"
+      },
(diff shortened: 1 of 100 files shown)


whitep4nth3r:
Hi @codenitrr, thank you for this suggestion!

After speaking with @erik-beus, I think it would make more sense to have each MDN link in a separate property for more flexibility on the docs (we can make a nice button rather than just including the link in the description, which could look repetitive), and as not to affect the formula description in the editor (and make it too long).

This would also need a JSON schema change. 

Is that something you could do?

whitep4nth3r:
Great! Thank you!

I think a specific property called `mdnUrl` would be useful, because we also want to provide a link to GitHub for the formulas and actions, as mentioned here https://github.com/nordcraftengine/documentation/issues/393

codenitrr:
@whitep4nth3r  Sorry, I had a few days off. I have now added the mdnUrl property. I had some trouble with git, so hopefully this pull request is okay, otherwise please let me know.

packages/lib/formulas/flatten/formula.json
```diff
@@ -2,16 +2,28 @@
   "$schema": "../../schemas/libFormula.schema.json",
   "name": "Flatten",
   "description": "Flatten a nested Array.",
+<<<<<<< HEAD
+=======
```
whitep4nth3r: You've got a few issues where the changes didn't merge correctly!

In your code editor, do a search for <<<< to find them all and fix them up.

packages/lib/formulas/greaterThan/formula.json
```diff
@@ -2,22 +2,36 @@
   "$schema": "../../schemas/libFormula.schema.json",
   "name": "Greater than",
   "description": "Compute if a value is larger than another value.",
+  "mdnUrl": "",
```
whitep4nth3r: We should make the mdnUrl optional in the schema, so formulas without mdnUrl links can just have the property omitted, rather than using an empty string.

whitep4nth3r:
@codenitrr did you commit the schema changes?

erik-beus:
@codenitrr I added some suggested changes to your branch here: https://github.com/codenitrr/nordcraft/pull/2
I thought they were easier to add in a PR since they also included some formatting.
Let me know if you have any concerns/questions. If not, I suggest you merge my PR, and then we can merge this PR afterwards :)

packages/lib/formulas/add/formula.json
```diff
@@ -5,20 +5,33 @@
...
-      "formula": { "type": "value", "value": 1 }
+      "formula": {
+        "type": "value",
+        "value": 1
+      }
     }
   ],
   "variableArguments": true,
   "description": "Get the sum of multiple numbers.",
+  "mdnUrl": "https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Add",
```
erik-beus:
```suggestion
  "mdnUrl": "https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Addition",
```

erik-beus (changes requested):
This link doesn't seem valid

whitep4nth3r:
Thanks @codenitrr!

PR #536, opened 2025-09-04
Allow auto fixing more rules

erik-beus:
Allow removing unused:
- APIs
- Attributes
- Events
- Component formulas

### DIFF
--- a/packages/search/src/rules/events/noReferenceEventRule.ts
+++ b/packages/search/src/rules/events/noReferenceEventRule.ts
@@ -41,7 +41,6 @@ export const noReferenceEventRule: Rule<{ name: string }> = {
     if (events.has(event.name)) {
       return
     }
-    console.log('Found issue', args.path, args.value.event.name)
     report(args.path, { name: args.value.event.name }, ['delete-event'])
   },
   fixes: {
--- a/packages/search/src/searchProject.ts
+++ b/packages/search/src/searchProject.ts
@@ -253,7 +253,6 @@ function* visitNode({
         // Report callback used to report issues
         (path, details, fixes) => {
           if (fixOptions) {
-            console.log('In fix mode', fixOptions.fixType, path, details, fixes)
             // We're in "fix mode"
             if (
               // We only overwrite fixedFiles once to avoid conflicting fixes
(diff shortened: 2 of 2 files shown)


packages/search/src/rules/apis/noReferenceApiRule.ts
```diff
@@ -1,42 +1,59 @@
...
-            formula.type === 'path' &&
-            formula.path[0] === 'Apis' &&
-            typeof formula.path[1] === 'string'
-          ) {
-            usedApis.add(formula.path[1])
-          }
+export type NoReferenceApiRuleFix = 'delete-api'
+
+const hasReferences = (args: NodeType) => {
+  if (args.nodeType !== 'component-api') {
```
jacob-kofoed: This is doing a bit more than `hasReferences`, but I understand this for the sake of avoiding code-duplication. I still think we should look into a way to do this in a standardized way - perhaps by running visit automatically before each fix, to make sure the same fix is reported.

packages/search/src/rules/apis/noReferenceApiRule.ts
```diff
@@ -1,42 +1,59 @@
...
+    return true
+  }
+  const { value, memo, component } = args
+  // eslint-disable-next-line @typescript-eslint/no-unnecessary-condition
+  if (!component) {
+    return true
+  }
+
+  const componentApiReferences = memo(
```
jacob-kofoed: How does memo work when fixing stuff? I suppose it is rarely actually memoized?

jacob-kofoed (approved):

packages/search/src/rules/apis/noReferenceApiRule.ts
erik-beus: When we fix issues across an entire project, I still think it matters

jacob-kofoed: but doesn't it reset after each fix? And if not, shouldn't it? To make sure something is still an issue?

erik-beus: Well, we no longer use it as part of the fix function now that we changed they way we run it.
Also, the memo function is reset every time we call `searchProject`, and we do this up to 100 times when we apply fixes. So for every fix, memos will be re-calculated.
Now since we only apply 1 fix at a time it probably doesn't matter, but I'm sure there's a slight overhead in memoizing things that we don't really need to memoize - especially in cases where we only fix 1 instance of an issue. But it should be okay I think 😉

PR #537, opened 2025-09-04
Add style validation issue rule

erik-beus:


packages/search/src/searchProject.ts
```diff
@@ -571,6 +576,20 @@ function* visitNode({
             })
           }
         }
+        yield* visitNode({
+          args: {
+            nodeType: 'style',
+            value: { style: value.style, element: value },
+            path: [...path, 'style'],
```
jacob-kofoed: You consider even more fine-grained check here by going into each style rule instead of the styles object. Same outcome, but the path would already be exact.

packages/core/src/utils/css.ts
```diff
@@ -0,0 +1,10 @@
+export const isCssPropertyNameValid = (property: string) => {
```
jacob-kofoed: You didn't end up using this? You plan on using it soon? In that case write a comment please.

packages/search/src/rules/style/invalidStyleSyntaxRule.ts
```diff
@@ -0,0 +1,28 @@
...
+}> = {
+  code: 'invalid style syntax',
+  level: 'error',
+  category: 'Quality',
+  visit: (report, { nodeType, value, path }) => {
+    if (nodeType !== 'style') {
+      return
+    }
+    for (const [prop, val] of Object.entries(value.style)) {
+      try {
```
jacob-kofoed: Dunno the cost of this, but you could potentially put this in a `memo` as the check is pure.

jacob-kofoed: The key would be the same `${prop}: ${val}`

jacob-kofoed: I wonder now if we should have a pure memo flag as well, as this wouldn't change from run to run 🤔

packages/core/src/utils/css.ts
erik-beus: I left it and forgot to remove it actually 🙈

jacob-kofoed (approved):
Super cool stuff!

PR #540, opened 2025-09-05
Modified structure for style-property definitions to live globally

jacob-kofoed:
Global custom style properties definitions are now living globally at the theme. Properties overrides now only has a formula.


packages/core/src/styling/customProperty.ts
```diff
@@ -28,29 +30,6 @@ export type CssSyntaxNode =
       keywords: string[]
     }
 
-/**
- * Default initial values for CSS properties based on their syntax type.
- *
- * Note: Eventually, the initial values may be set in a theme or somewhere global.
- */
-const DEFAULT_INITIAL_VALUE_BY_SYNTAX: Record<CssSyntax, string> = {
```
jacob-kofoed, line 36: We should instead enforce an initial value from the editor. It is always required except when syntax is "`*`"

packages/search/src/rules/style-variables/ambiguousStyleVariableSyntaxRule.ts
jacob-kofoed, line 1: This issue rule is no longer needed, but we will likely need some other ones.

erik-beus (approved):
Looks good 🔥

PR #545, opened 2025-09-08
Issue rule to reduce static formulas

jacob-kofoed:
The whole static check is interesting in regards to compiler or pre-processing as well. It doesn't work with core formulas yet, but we would need to specifically flag each formula that is **pure**. For pure formulas, we should be able to run apply formula without any context like Variable or environment data.

Dunno if this idea is the way to go, but perhaps until we dive further down the rabbit hole this can help flag some project issues.

### DIFF
--- a/packages/search/src/util/contextlessEvaluateFormula.ts
+++ b/packages/search/src/util/contextlessEvaluateFormula.ts
@@ -1,0 +1,111 @@
+import { type Formula } from '@nordcraft/core/dist/formula/formula'
+
+/**
+ * Static evaluation of a formula.
+ *
+ * Can be used by issues to determine if a formula or sub-formula can be reduced to a static value.
+ * When sophisticated enough, it can be used during compile-time to reduce all static subgraphs of a formula to static values, greatly reducing payload and improving runtime performance.
+ *
+ * @returns {
+ *  isStatic: boolean; // Whether the formula is static (i.e., does not depend on any variables, context AND only use pure formulas (no Random, Date, etc.))
+ *  result: unknown; // The evaluated value of the formula
+ * }
+ *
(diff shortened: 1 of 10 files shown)


packages/search/src/util/contextlessEvaluateFormula.ts
```diff
@@ -0,0 +1,111 @@
...
+ * Can be used by issues to determine if a formula or sub-formula can be reduced to a static value.
+ * When sophisticated enough, it can be used during compile-time to reduce all static subgraphs of a formula to static values, greatly reducing payload and improving runtime performance.
+ *
+ * @returns {
+ *  isStatic: boolean; // Whether the formula is static (i.e., does not depend on any variables, context AND only use pure formulas (no Random, Date, etc.))
+ *  result: unknown; // The evaluated value of the formula
+ * }
+ *
+ * TODO: Make this function more capable of evaluating pure core formulas.
+ * TODO: Memorize the results (using path or a fast hash) to avoid re-evaluating any similar sub-graphs multiple times.
```
erik-beus:
```suggestion
 * TODO: Memoize the results (using path or a fast hash) to avoid re-evaluating any similar sub-graphs multiple times.
```

packages/search/src/util/contextlessEvaluateFormula.ts
```diff
@@ -0,0 +1,111 @@
...
+ * TODO: Add a complex test-suite to ensure it works and develops as expected.
+ */
+
```
erik-beus:
```suggestion
```

packages/search/src/util/contextlessEvaluateFormula.ts
jacob-kofoed: Corected the the gramar as suggessted

PR #550, opened 2025-09-10
Workflow callbacks

jacob-kofoed:
I had this branch laying around for a while and seems to be working. Requires the "`wfc`" editor branch to be merged and the "`Workflow callbacks`" feature-flag to be set to test.

### DIFF
--- a/packages/search/src/types.d.ts
+++ b/packages/search/src/types.d.ts
@@ -219,6 +219,10 @@ type ComponentWorkflowNode = {
           testValue: any
         }[]
       | null
+    callbacks?: {
+      name: string
+      testValue: any
+    }[]
     actions: ActionModel[]
     exposeInContext?: boolean | undefined
   }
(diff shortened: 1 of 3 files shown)


packages/search/src/types.d.ts
```diff
@@ -219,6 +219,10 @@ type ComponentWorkflowNode = {
           testValue: any
         }[]
       | null
+    callbacks?: {
+      name: string
+      testValue: any
```
erik-beus, line 224: We should make sure to remove this in the `removeTestData` function 👀 https://github.com/nordcraftengine/nordcraft/blob/main/packages/ssr/src/rendering/testData.ts#L10

erik-beus (approved):
Looks good 👍 
We should also remember to add issue rules for invalid callback names btw. But that could be added later

packages/search/src/types.d.ts
jacob-kofoed: Good point. Done! I've added a test-case for it as well.

erik-beus:
@jacob-kofoed you should also make sure to update `getFormulasInAction` in our `formulaUtils` probably? To make sure we visit formulas/actions during search in our problems worker?

PR #551, opened 2025-09-10
Adjust search worker to include an id in messages

erik-beus:
Including an id in messages (both requests and responses) allows the main thread to match responses to requests, which is especially useful when multiple requests are made concurrently. In our case, we both request problems for a single file and for the entire project.

### DIFF
--- a/package.json
+++ b/package.json
@@ -2,7 +2,7 @@
   "name": "nordcraft",
   "type": "module",
   "license": "Apache-2.0",
-  "version": "1.0.45",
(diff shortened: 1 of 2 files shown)


jacob-kofoed (approved):
LGTM. Dunno if somehow useable here, but there is a native ID generated for message requests https://developer.mozilla.org/en-US/docs/Web/API/MessageEvent/lastEventId

Not sure how the sender knows of this Id though 🤔

PR #553, opened 2025-09-11
Omit to splice arrays

jacob-kofoed:
This changes how our omit function works by actually removing items from arrays instead of replacing them with empty nodes. The omit function is not used that much, so this isn't a major change.

You can test if the quick-fix of removing unused events works now and doesn't leave empty-nodes behind (which would crash the problems.worker)

### DIFF
--- a/packages/core/src/utils/collections.ts
+++ b/packages/core/src/utils/collections.ts
@@ -32,7 +32,7 @@ export const omit = <T = object>(
   }
 
   if (Array.isArray(collection)) {
-    return collection.toSpliced(key as number, 1) as T
+    return collection.toSpliced(Number(key), 1) as T
   }
 
   const clone: any = { ...collection }


packages/core/src/utils/collections.ts
```diff
@@ -19,24 +19,25 @@ export const mapValues = <T, T2>(
...
+  [key, ...rest]: Array<string | number>,
 ): T => {
-  const [head, ...rest] = path
-
-  const clone: any = Array.isArray(collection)
-    ? [...collection]
-    : isObject(collection)
-      ? { ...collection }
-      : {}
+  if (rest.length) {
```
erik-beus: I would prefer if we were specific here 🙏 

```suggestion
  if (rest.length > 0) {
```

packages/core/src/utils/collections.ts
```diff
@@ -19,24 +19,25 @@ export const mapValues = <T, T2>(
...
+    clone[key] = omit(clone[key], rest)
+    return clone
+  }
 
-  if (rest.length === 0) {
-    delete clone[head]
-  } else {
-    clone[head] = omit(clone[head], rest)
+  if (Array.isArray(collection)) {
+    return collection.toSpliced(key as number, 1) as T
```
erik-beus: Consider casting this to a number since it might be a string?

```suggestion
    return collection.toSpliced(Number(key), 1) as T
```

PR #554, opened 2025-09-11
Allow deleting unknown attributes with autofix

erik-beus:


packages/search/src/rules/attributes/unknownAttributeRule.test.ts
```diff
@@ -91,3 +97,61 @@ describe('unknownAttribute', () => {
...
+              attrs: {
+                unknown: {
+                  type: 'path',
+                  path: ['Attributes', 'unknown'],
+                },
+              },
+              children: [],
+              events: {},
+              style: {},
+              name: 'unknown-component',
```
jacob-kofoed, line 135: Maybe have a test where this component exists, but doesn't have the attribute? Also, consider giving it two or more attributes and make sure only exactly the unknown one is removed. Would potentially have caught the issue with saw with `Omit` etc.

erik-beus: Good suggestion. This actually highlighted a number of issues with this rule's fix. I'll get those fixed next week 👍

erik-beus:
I'll close this, as there is no clear fix to when you reference an unknown attribute (within a component).

PR #564, opened 2025-09-18
Fixed a bug where videos responses would be parsed as text

ChrEsb:
If an API requests a video file, that file would be parsed as text. This is a very heavy operation and would slow performance down to an unusable level. This PR will ensure we automatically parse videos as blobs instead.


ChrEsb:
This PR is a breaking change. We'll wait for versioning to roll this out

PR #569, opened 2025-09-23
Add relevant MDN links as comments in lib formulas

erik-beus:
- To clarify that the standard formulas are based on web standards, we're adding links to the MDN documentation for most formulas
- Adjust validation schema for formulas
- Remove obsolete code

### DIFF
--- a/packages/lib/formulas/reduce/handler.ts
+++ b/packages/lib/formulas/reduce/handler.ts
@@ -2,7 +2,7 @@ import type { FormulaHandler } from '@nordcraft/core/dist/types'
 
 /**
  * Similar to https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/reduce
- * but also works with objects.
+ * but also works with objects
  */
 const handler: FormulaHandler<unknown> = ([items, fx, init]) => {
   if (typeof fx !== 'function') {


packages/lib/formulas/fromEntries/handler.ts
```diff
@@ -1,5 +1,9 @@
 import type { FormulaHandler } from '@nordcraft/core/dist/types'
 
+/**
+ * Similar to https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/fromEntries
+ * but requires the input to be an array of objects with `key` and `value` properties.
```
whitep4nth3r:
```suggestion
 * but requires the input to be an array of objects with `key` and `value` properties named as `key` and `value` respectively
```

packages/lib/formulas/reduce/handler.ts
```diff
@@ -1,8 +1,11 @@
 import type { FormulaHandler } from '@nordcraft/core/dist/types'
 
+/**
+ * Similar to https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/reduce
+ * but also works with objects.
```
whitep4nth3r:
```suggestion
 * but also works with objects
```

PR #574, opened 2025-09-24
Handle multiple themes with custom properties

jacob-kofoed:
We previously expected exactly one theme, even though they were stored in a Record. This change is a first step towards using multiple themes.

Breaking-change (Requires update for rendering): https://github.com/nordcraftengine/nordcraft-internal/pull/3270

### DIFF
--- a/packages/runtime/src/editor-preview.main.ts
+++ b/packages/runtime/src/editor-preview.main.ts
@@ -1104,10 +1104,13 @@ export const createRoot = (
                   .filter(([key]) => previewStyleStyles[key])
                   .map(([key, val]) => [
                     key,
-                    { ...val, initialValue: previewStyleStyles[key] },
+                    { ...val, value: previewStyleStyles[key] },
                   ]),
               )
               const cssBlocks: string[] = []
+              if (theme.value.default) {
+                cssBlocks.push(renderTheme(`:host, :root`, theme.value))
+              }
               if (theme.value.defaultDark) {
                 cssBlocks.push(
                   renderTheme(


packages/core/src/styling/customProperty.ts
```diff
@@ -41,18 +41,72 @@ export function stringifySyntaxNode(node: CssSyntaxNode): string {
...
+  while ((match = VAR_REGEX.exec(initialValue))) {
+    const varName = match[1]
+    if (
+      !isDefined(theme.propertyDefinitions?.[varName as CustomPropertyName])
+    ) {
+      return null
+    }
+
+    const value =
+      theme.propertyDefinitions?.[varName as CustomPropertyName].initialValue
```
erik-beus: I don't think you need optional chaining here if you you extract `theme.propertyDefinitions?.[varName as CustomPropertyName]` to a const above (before the `if` statement)

packages/core/src/styling/theme.ts
```diff
@@ -71,49 +71,85 @@ export type OldTheme = {
   shadow: Record<string, { value: string; order: number }>
   breakpoints: Record<string, { value: number; order: number }>
 }
+
 export type Theme = {
+  default?: true
+  defaultDark?: true
+  defaultLight?: true
+  propertyDefinitions?: Record<CustomPropertyName, CustomPropertyDefinition>
+
```
erik-beus:
```suggestion
```

packages/core/src/styling/theme.ts
```diff
@@ -71,49 +71,85 @@ export type OldTheme = {
...
+        defaultTheme,
+      ),
     )
     .join('\n')}
+
+  /* Render default theme */
+  ${renderTheme(':host, :root', defaultTheme)}
+
+  ${renderTheme(':host, :root', defaultDarkTheme, '@media (prefers-color-scheme: dark)')}
+  ${renderTheme(':host, :root', defaultLightTheme, '@media (prefers-color-scheme: light)')}
```
erik-beus, line 137: Consider creating a function that renders multiple themes. Then, if no dark/light theme exist, that function could make sure we don't add unnecessary newlines.

packages/core/src/styling/theme.ts
```diff
@@ -353,3 +405,33 @@ ${RESET_STYLES}
   }
 }`
 }
+
+export function renderTheme(
+  selector: string,
+  theme: Theme | undefined,
+  mediaQuery?: string,
+) {
+  const properties = Object.entries(theme?.propertyDefinitions ?? {}).filter(
```
erik-beus: Consider returning earlier if `!theme` to avoid the `.filter` call in that case

packages/ssr/src/rendering/head.ts
```diff
@@ -62,19 +62,19 @@ export const getHeadItems = ({
...
-    // We include all fonts even though it's not necessary.
-    // While this is not the ideal long-term solution, it does have a few benefits:
-    // - It's easier to cache the font stylesheet across pages
-    // - It simplifies style variable setup in theme.ts
-    // - It ensures the same behaviour as in our editor
-    // - It increases the chance that there's a font to be used by our
-    //   reset stylesheet (font-family: var(--font-sans))
-    // For most apps, the overhead of including all fonts is negligible and
-    // it doesn't add a lot of bytes to our stylesheet.
+  Object.values(themes).forEach((theme) => {
```
erik-beus, line 65: Ideally, we should make sure not to add the same font twice if it's defined in multiple themes. I guess it's an existing issue though since it could exist multiple times in the same theme as well 🤷‍♂️

erik-beus (changes requested):
Please see suggestions

packages/ssr/src/rendering/head.ts
jacob-kofoed: I put them in a `Set` and construct them afterwards.

erik-beus (approved):

PR #577, opened 2025-09-24
Improve test data removal and visit more formulas

erik-beus:
This reduces project sizes with ~2,5%


packages/core/src/component/component.types.ts
```diff
@@ -217,6 +217,7 @@ export interface ComponentWorkflow extends NordcraftMetadata {
   callbacks?: Array<{ name: string; testValue: any }>
   actions: ActionModel[]
   exposeInContext?: boolean
+  testValue?: any
```
jacob-kofoed: Looks safe to use `unknown` instead of any here? Otherwise put in the actual possible types. `Array<unknown> | Object | String | Number | null`

jacob-kofoed (approved):

PR #581, opened 2025-09-29
Rearrange issue rules and enable 2 more rules

erik-beus:
- Restructure all rules to make it easier to make sure all rules are included.
- Enable the `noReferenceApiInputRule` and `unknownApiInputRule`


bin/prepareReleaseAssets.ts
```diff
@@ -99,6 +101,31 @@ const run = async () => {
...
   )
 
+  // Build issue rules list
+  createTempFileFromValue(
+    'issue-rules.json',
+    JSON.stringify(
+      {
+        rules: groupBy(
+          easySort(
+            ISSUE_RULES.map((r) => ({
```
jacob-kofoed: Would be nice with a helper `pick` function for stuff like this:

```ts
.map(r => pick(r, ["category", "code", "level"...]))
```

Similar to https://www.geeksforgeeks.org/javascript/lodash-_-pick-method/

packages/search/src/rules/issues/apis/noReferenceApiInputRule.ts
```diff
@@ -1,5 +1,5 @@
 import { ToddleApiV2 } from '@nordcraft/core/dist/api/ToddleApiV2'
-import type { Rule } from '../../types'
+import type { Rule } from '../../../types'
```
erik-beus, line 2: This rule was not enabled/included before

packages/search/src/rules/issues/apis/unknownApiInputRule.ts
```diff
@@ -1,5 +1,5 @@
 import type { ApiRequest } from '@nordcraft/core/dist/api/apiTypes'
-import type { Rule } from '../../types'
+import type { Rule } from '../../../types'
```
erik-beus, line 2: This rule was not enabled/included before

PR #592, opened 2025-09-30
Store and restore scroll-position per component

jacob-kofoed:
Most browsers would already remember scroll-position on the window element when editing the document (iframe), but would not remember other scroll containers, which seems to be a common pattern. This task scope creeped a bit to include a few more features. Please test:

- Keep scroll position of the window when modifying a component.
- Keep scroll positions when going between different components and back again.
- Keep scroll position even after the canvas is closed down, ie. going to a global action and then back to a component.
- Keep all scroll-position on reload (within the same session).
- Does not keep scroll-position between tabs.

All of these features are similar to how vs-code etc. handles the view-state.

Solves [#3268](https://github.com/nordcraftengine/nordcraft-internal/issues/3268)


packages/runtime/src/utils/storeScrollState.ts
```diff
@@ -0,0 +1,65 @@
+export const storeScrollState = (
+  key: string = '',
```
erik-beus, line 11: Why is the key optional?

packages/runtime/src/utils/storeScrollState.ts
```diff
@@ -0,0 +1,65 @@
...
+  querySelector = '[data-id]',
+  comparerFn = (node: Element) => node.getAttribute('data-id'),
+) => {
+  const scrollPositions: Record<
+    string,
+    {
+      x: number
+      y: number
+    }
+  > = {}
```
erik-beus: Consider extracting a type for the `scrollPositions` to avoid repeating it multiple times

erik-beus (changes requested):
Please see questions

packages/runtime/src/utils/storeScrollState.ts
jacob-kofoed: In cases where I use the curry function instead of looking it up again in session-storage from a key, it is technically not required.

packages/runtime/src/utils/storeScrollState.ts
```diff
@@ -0,0 +1,64 @@
+type ScrollPosition = {
+  x: number
+  y: number
+}
```
erik-beus: Consider this to get better auto completion + highlight that every key doesn't necessarily have a value:

```suggestion
interface ScrollPosition {
  x: number
  y: number
}

type ScrollPositionKey = '__window' | string

type ScrollPositions = Partial<Record<ScrollPositionKey, ScrollPosition>>
```

erik-beus (approved):
It works as expected 👍

PR #606, opened 2025-10-08
Cleanup types for style variants

erik-beus:
We had 2 declarations/versions of the types for style variants. This cleans it up to use only one as well as simplify the types for style variant styles.
This cleanup also means we no longer need to rely on types from `react`.

### DIFF
--- a/packages/core/src/styling/style.css.ts
+++ b/packages/core/src/styling/style.css.ts
@@ -1,15 +1,15 @@
 // cSpell:ignore thinn, ABCDEFGHIJKLMNOPQRSTYVWXYZ
-import type { Component } from '../component/component.types'
+import type {
+  Component,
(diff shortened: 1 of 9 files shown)


packages/core/src/styling/style.css.ts
```diff
@@ -136,10 +136,10 @@ export const createStylesheet = (
         'breakpoints',
         'shadows',
       ])
-      const styleVariants = node.variants ?? node.style?.variants
```
erik-beus, line 139: This bit is technically a breaking change, but I couldn't see how/why we would've ever kept the style variants in the style object on a node? 🤔

jacob-kofoed: Not sure I would dare change this before versioning. Editor project seems to work though

PR #712, opened 2025-12-10
Use MDN's element descriptions

erik-beus:
Rather than using our AI generated element descriptions, we should use the descriptions from MDN, which are more accurate and reliable. These descriptions will be used in the element catalog in the editor.


mikkelnordcraft (approved):

erik-beus:
We'll pause this for now until the element catalog better supports longer descriptions 👍

PR #839, opened 2026-04-07
Make API handling more robust

erik-beus:
In case an API was set to null, removing test data from a component would not work among other things.
This PR makes sure we treat APIs as nullish to better support these edge cases.


jacob-kofoed (approved):
Looks good. Classic problem at this point.

PR #843, opened 2026-04-13
Better handle multiple custom-elements on the same page, when using the new theming system

jacob-kofoed:
There was an issue where custom elements would reference the wrong stylesheet when multiple web components were used on the same website.

### DIFF
--- a/packages/runtime/src/utils/subscribeCustomProperty.ts
+++ b/packages/runtime/src/utils/subscribeCustomProperty.ts
@@ -4,7 +4,10 @@ import { CUSTOM_PROPERTIES_STYLESHEET_ID } from '@nordcraft/core/dist/styling/th
 import type { StyleVariant } from '@nordcraft/core/dist/styling/variantSelector'
 import { CustomPropertyStyleSheet } from '../styles/CustomPropertyStyleSheet'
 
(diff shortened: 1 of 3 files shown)


packages/runtime/src/utils/subscribeCustomProperty.ts
```diff
@@ -4,7 +4,10 @@ import { CUSTOM_PROPERTIES_STYLESHEET_ID } from '@nordcraft/core/dist/styling/th
 import type { StyleVariant } from '@nordcraft/core/dist/styling/variantSelector'
 import { CustomPropertyStyleSheet } from '../styles/CustomPropertyStyleSheet'
 
-export let customPropertiesStylesheet: CustomPropertyStyleSheet | undefined
+export const customPropertiesStylesheets = new WeakMap<
```
jacob-kofoed, line 7: Main change really. Instead of a singleton for the stylesheet, it is now one for each root (document or shadow-root).

I am honestly a little puzzled why this was happening, as I thought the scope was per component, as they each import the script, not using anything outside their shadow-root. Apparently <script type="module"> inside shadow DOM is shared across all instances if URL is exact match, not per component.

Good learning, and I suppose this makes them less RAM hungry, so it does make sense that they run in the same instance.

erik-beus (approved):

PR #851, opened 2026-04-17
Ensure workflow events are visited by issue rules

erik-beus:
Actions/formulas defined in workflow event callbacks were not visited by issue/search rules. This PR:

- Ensures we visit workflow event callbacks when running issue/search rules
- Relaxes types for workflow callbacks to safeguard agains nullish values
  
Closes https://github.com/nordcraftengine/nordcraft-internal/issues/4158


jacob-kofoed (approved):
lgtm – looks like I was the one who was the one who forgot the callback actions here when I made it. Did you check that it is added in the ToddleCompoment search we use for which actions to include in custom code as well?

PR #854, opened 2026-04-17
Report should fallback to standard branch eval when reaching a max-depth

jacob-kofoed:


packages/runtime/src/editor-preview.main.ts
```diff
@@ -251,22 +251,35 @@ export const createRoot = (
...
+        componentFormulaData[path.join('/')] = JSON.parse(JSON.stringify(data))
+      } catch {
+        componentFormulaData[path.join('/')] =
+          `[Unserializable value of type ${typeof data}]`
+      } finally {
+        reportComponentFormulaData()
+      }
+    }
   }
-  const reportComponentFormulaData = debounce(
```
jacob-kofoed: This debounce was a major performance issue as clearTimeout & setTimeout has a lot of overhead, and we would call it thousands of times. The throttle implementation below waits for first next idle time and only sets up one callback per batch.

packages/runtime/src/components/createComponent.ts
```diff
@@ -94,11 +98,15 @@ export function createComponent({
...
-            })
+            applyFormula(
+              api!.autoFetch,
+              {
+                ...formulaCtx,
+                component,
+                data: dataSignal.get(),
+              },
+              ['apis', name, 'autoFetch'],
+            ) === true
```
erik-beus: Consider comparing using `toBoolean` instead? 🤔

erik-beus (approved):

PR #855, opened 2026-04-17
Add css-parser package used in Nordcraft editor

erik-beus:
Open source our internal CSS parser package, which is used in the Nordcraft editor to power the style panel.


marija04 (approved):
Looks good 💪

PR #917, opened 2026-05-04
Fix the issue when there is an invalid value in a background

marija04:
There was an issue when there is an invalid value in a background gradient we will set that to many properties instead of just one. For example here 'linear-gradient(wjote, #fff200 0% 100%, white 50%, white 55%, yellow 100%) scroll border-box padding-box repeat repeat 0% 0% / auto'

We will set "wjote" as a direction and interpolation and we want only one of them and that will be direction in this case.


mikkelnordcraft (approved):
🚀

PR #926, opened 2026-05-05
More accurately display text when editing in canvas

jacob-kofoed:
Also fix an issue where setting component instance styling on a page or component that starts with a number (most commonly page `404`) would not work.

Also improve the performance measurer with colors to differentiate renders and formulas.

### DIFF
--- a/packages/core/src/utils/getNodeSelector.ts
+++ b/packages/core/src/utils/getNodeSelector.ts
@@ -19,7 +19,10 @@ export function getNodeSelector(
 ): string {
   let selector = `[data-id="${path}"]`
   if (componentName) {
-    selector += `.${componentName}`
+    // Do not allow classes to start with a number, for example a page named "404" would result in a selector starting with a number which is invalid in CSS.
+    selector += startsWithNumber(componentName)
+      ? `._${componentName}`
+      : `.${componentName}`
   }
   if (nodeId) {
     selector += `\\:${nodeId}`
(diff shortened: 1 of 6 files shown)


packages/core/src/utils/measure.ts
```diff
@@ -1,14 +1,20 @@
 const globalScope: any = typeof globalThis !== 'undefined' ? globalThis : window
-globalScope.__nc_measure_max_depth = 10
+globalScope.__nc_measure_max_depth =
+  typeof sessionStorage !== 'undefined' &&
+  sessionStorage.getItem('__nc_measure_max_depth')
+    ? parseInt(sessionStorage.getItem('__nc_measure_max_depth')!)
```
erik-beus, line 5: Not sure if esbuild will optimize this, but instead of reading twice from sessionStorage, you could save the result in a const

packages/core/src/utils/getNodeSelector.ts
```diff
@@ -19,7 +19,10 @@ export function getNodeSelector(
 ): string {
   let selector = `[data-id="${path}"]`
   if (componentName) {
-    selector += `.${componentName}`
+    // Do not allow classes to start with a number, for example a page named "404" would result in a selector starting with a number which is invalid in CSS.
+    selector += startsWithNumber(componentName)
```
erik-beus, line 23: This would also affect how classnames are generated during SSR right?

erik-beus (approved):

marija04:
@jacob-kofoed I have made some changes that are already merged but haven't bump the version yet. Can you maybe do it with this PR?

PR #933, opened 2026-05-06
Update the bun version

marija04:


.github/workflows/deploy_css_parser.yml
erik-beus, line 9: You might want to add this workflow file to the list of paths that should trigger a deploy of the CSS parser? Otherwise, this change will not trigger a new deploy.

```suggestion
      - 'packages/css-parser/**'
      - '.github/workflows/deploy_css_parser.yml'
```

.github/workflows/deploy_css_parser.yml
```diff
@@ -17,7 +17,7 @@ jobs:
       - name: Install bun
         uses: oven-sh/setup-bun@v2
         with:
-          bun-version: 1.3.3
+          bun-version: 1.3.13
```
erik-beus, line 21: I suggest you also bump it in this workflow file: `.github/workflows/test_css_parser.yml`
Alternatively across the entire repository 🤞

erik-beus (changes requested):

erik-beus (approved):

PR #935, opened 2026-05-06
Update action/checkout to v6

marija04:


erik-beus (changes requested):
~Could we try it like this before merging?~

erik-beus (approved):

PR #959, opened 2026-05-08
Change API redirect statusCode type to formula

erik-beus:
To allow more flexibility, the redirect status code should be a formula instead of a number. Since the `statusCode` property has never been used, it should be safe to change its type. Also - even if it had previously been set to a `number`, `applyFormula` would still return a number.

### DIFF
--- a/packages/backend/src/utils/api.ts
+++ b/packages/backend/src/utils/api.ts
@@ -5,13 +5,14 @@ import {
   requestHash,
   sortApiEntries,
 } from '@nordcraft/core/dist/api/api'
-import type {
-  ApiMethod,
-  ApiParserMode,
-  ApiPerformance,
-  ApiStatus,
-  LegacyApiStatus,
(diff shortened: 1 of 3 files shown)


packages/backend/src/utils/api.ts
```diff
@@ -297,26 +296,32 @@ const fetchApiV2 = async ({
...
       const url = validateUrl({
         path: location,
         origin: originalRequestUrl.origin,
       })
       if (url) {
         // Opt out early to avoid additional API requests/rendering
+        let statusCode = applyFormula(rule.statusCode, ruleContext)
+        statusCode =
+          typeof statusCode === 'number' &&
+          REDIRECT_STATUS_CODES.includes(statusCode as RedirectStatusCode)
```
mikkelnordcraft: I hate that include does this 👿

mikkelnordcraft (approved):

PR #961, opened 2026-05-08
Using calc in animation

### LINKED ISSUE #918: Animation editor does not accept css functions in duration
Whe trying to use e.g. a calc function in duration it reverts to 0s.

Seems to work with other inputs in the animation editor

### LINKED ISSUE #962: Using undefined variable for border-color
See the discussion [here](https://discord.com/channels/972416966683926538/1501675247060123759)

marija04:
We want to support css function for the animation duration and animation delay properties. Verify you are able to use the calc, clamp, min and max function for these two properties.

In the same PR I'm also fixing an issue with the border. If you set a border-color to a variable that is not defined we still want to show it in the style panel so the user can update it. This issue was reported [here](https://discord.com/channels/972416966683926538/1501675247060123759)

### DIFF
--- a/packages/css-parser/src/parseAnimation.ts
+++ b/packages/css-parser/src/parseAnimation.ts
@@ -659,12 +659,11 @@ const isValidTimeValue = ({
       const val = parseMultipleValuesRes[i]
       if (validValue && isDefined(val)) {
         if (val.type === 'function' && CSS_FUNCTIONS.includes(val.name)) {
-          const isValid = isValidTimeValue({
+          validValue = isValidTimeValue({
             valueToCheck: val.value,
             variables,
             isCalc: val.name === 'calc',
           })
-          validValue = isValid
         } else if (val.type === 'function' && val.name === 'var') {
           // If it's a variable
           const allValues = val.value.split(',').map((v) => v.trim())
--- a/packages/css-parser/src/parseAnimation.ts
+++ b/packages/css-parser/src/parseAnimation.ts
@@ -457,7 +457,7 @@ export const getParsedAnimation = (
   }
 }
 
-const cssFunctions = ['calc', 'clamp', 'max', 'min']
+const CSS_FUNCTIONS = ['calc', 'clamp', 'max', 'min']
 
 const parseAnimation = ({
   valueToCheck,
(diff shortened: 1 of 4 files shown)


packages/css-parser/src/parseAnimation.ts
```diff
@@ -457,6 +457,8 @@ export const getParsedAnimation = (
   }
 }
 
+const cssFunctions = ['calc', 'clamp', 'max', 'min']
```
erik-beus: Consider making it clear that this is a constant

```suggestion
const CSS_FUNCTIONS = ['calc', 'clamp', 'max', 'min']
```

packages/css-parser/src/parseAnimation.ts
```diff
@@ -617,3 +633,96 @@ const parseAnimation = ({
     invalidValue,
   }
 }
+
+const checkIfTimeValues = ({
```
erik-beus: Consider renaming to `isValidTimeValue`

packages/css-parser/src/parseAnimation.ts
```diff
@@ -617,3 +633,96 @@ const parseAnimation = ({
...
+        if (val.type === 'function' && cssFunctions.includes(val.name)) {
+          const t2 = checkIfTimeValues({
+            valueToCheck: val.value,
+            variables,
+            isCalc: val.name === 'calc',
+          })
+          validValue = t2
+        } else if (val.type === 'function' && val.name === 'var') {
+          // If it's a variable
+          const allValues = val.value.split(', ')
```
erik-beus: Will there always be a trailing space after the comma? Alternatively:

```suggestion
          const allValues = val.value.split(',').map(v => v.trim())
```

erik-beus (changes requested):
Please see suggestions

packages/css-parser/src/parseAnimation.ts
```diff
@@ -617,3 +633,96 @@ const parseAnimation = ({
...
+    ) {
+      const val = parseMultipleValuesRes[i]
+      if (validValue && isDefined(val)) {
+        if (val.type === 'function' && CSS_FUNCTIONS.includes(val.name)) {
+          const isValid = isValidTimeValue({
+            valueToCheck: val.value,
+            variables,
+            isCalc: val.name === 'calc',
+          })
+          validValue = isValid
```
erik-beus:
```suggestion
          validValue = isValidTimeValue({
            valueToCheck: val.value,
            variables,
            isCalc: val.name === 'calc',
          })
```

erik-beus (approved):
Final suggestion :)

PR #966, opened 2026-05-11
More context issue rules (re-revert)

jacob-kofoed:


packages/search/src/rules/issues/context/noReferenceContextWorkflowRule.ts
```diff
@@ -0,0 +1,89 @@
...
+  visit: (report, { path, files, value, nodeType, memo }) => {
+    if ((nodeType as NodeType['nodeType']) !== 'component-context') {
+      return
+    }
+
+    if (Object.keys(value.workflows).length === 0) {
+      return
+    }
+
+    const [_fileType, componentName, _contexts, contextComponentName] =
```
jacob-kofoed, line 27: I believe this fixes the false-positives we saw earlier. I used the "componentName" on the context subscription, which in some (probably legacy) cases was undefined. I now use the key from the path instead which is guaranteed to be given.

erik-beus (approved):

PR #1007, opened 2026-05-13
Visit API redirect rule status code formulas

erik-beus:
- Ensure formulas for status code for API redirect rules are visited when we find all formula references
- Add tests to ensure this is working correctly
- Also test that the `enabled` formula on meta tags is visited when we find all formula references
- Expose a type


marija04 (approved):
Looks good ✅

PR #1018, opened 2026-05-15
Reduce serialization overhead by stripping styles

erik-beus:
- Stop serializing `style` and `variants` but replace them with a static classname for all nodes
- Only add classname (client side) on elements that actually have any styling/variants. This means that styling in the editor will be unaffected
- Cleanup types

For the editor project, this reduces the amount of data we need to serialize by ~5%


packages/ssr/src/rendering/classes.ts
```diff
@@ -0,0 +1,42 @@
...
+
+/**
+ * Function to strip styles and variants from a Component's nodes and convert them to static class names
+ */
+export const serializeClasses = (component: Component) => ({
+  ...component,
+  nodes: mapObject<NodeModel, NodeModel>(
+    component.nodes ?? {},
+    ([key, node]) => {
+      if (node.type !== 'element') {
```
jacob-kofoed: Maybe we could do something similar to instance styling on component type nodes 🤔

packages/ssr/src/rendering/classes.ts
```diff
@@ -0,0 +1,42 @@
...
+  NodeModel,
+} from '@nordcraft/core/dist/component/component.types'
+import { valueFormula } from '@nordcraft/core/dist/formula/formulaUtils'
+import { getClassName } from '@nordcraft/core/dist/styling/className'
+import { mapObject } from '@nordcraft/core/dist/utils/collections'
+
+/**
+ * Function to strip styles and variants from a Component's nodes and convert them to static class names
+ */
+export const serializeClasses = (component: Component) => ({
```
jacob-kofoed: Not certain this counts as "serialization", maybe "resolveClasses" is a better name. Opininated though 🙊

jacob-kofoed (approved):
Initial test looks good, but I would like to test a bit more the release candidate version.

packages/ssr/src/rendering/classes.ts
erik-beus: You're right. However, it's a bit more tricky since we need to forward instance styling on wrapping components. Also, the amount of instance styling would probably take up far less space in most projects.
But it's definitely worth doing at some point 🤞

packages/ssr/src/rendering/classes.ts
erik-beus: Good point. I've updated the name now 👍

jacob-kofoed (approved):

PR #1028, opened 2026-05-18
Include API dependency info in SSR serialization

### LINKED ISSUE #1023: Sort APIs on server instead of client
APIs needs to run in order if they depend on each other. This is done both server side and client side. Performance improvement of sending the calculated dependency graph to the client could be noticable as all formulas has to be checked, but more importantly the bundle size can be greatly reduced if we don't need to reference:
- `sortApiObjects`
- `isLegacyApi`

Image

erik-beus:
- Added `dependsOn` property to the serialized API objects in the SSR rendering process, allowing the client to sort APIs without expensive calculations
- Adjust types and use the `ToddleComponent` type less when evaluating APIs during SSR
- Cleanup types

This reduces the runtime (`page.main.js`) from:
32.5 KB (114KB)
to:
30.5 KB (101KB)

That's a ~6% reduction (11.5%)

Closes https://github.com/nordcraftengine/nordcraft/issues/1023

### DIFF
--- a/packages/core/src/api/api.ts
+++ b/packages/core/src/api/api.ts
@@ -15,7 +15,7 @@ import type {
 import { ApiMethod } from './apiTypes'
 import { isJsonHeader } from './headers'
 import { LegacyToddleApi } from './LegacyToddleApi'
-import { ToddleApiV2 } from './ToddleApiV2'
+import type { ToddleApiV2 } from './ToddleApiV2'
 
 export const NON_BODY_RESPONSE_CODES = [101, 204, 205, 304]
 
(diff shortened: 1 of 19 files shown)


packages/core/src/api/api.ts
```diff
@@ -14,15 +14,14 @@ import type {
...
+import type { LegacyToddleApi } from './LegacyToddleApi'
+import type { ToddleApiV2 } from './ToddleApiV2'
 
 export const NON_BODY_RESPONSE_CODES = [101, 204, 205, 304]
 
 export const isLegacyApi = <Handler>(
   api: ComponentAPI | LegacyToddleApi<Handler> | ToddleApiV2<Handler>,
-): api is LegacyComponentAPI | LegacyToddleApi<Handler> =>
-  api instanceof LegacyToddleApi ? true : !('version' in api)
+): api is LegacyComponentAPI | LegacyToddleApi<Handler> => !('version' in api)
```
erik-beus, line 24: This still works even if the input is a class. See previous PR that fixed an issue when using `Object.hasOwn` here https://github.com/nordcraftengine/nordcraft-internal/pull/1189

packages/ssr/src/rendering/api.ts
```diff
@@ -19,3 +25,61 @@ export type ApiEvaluator = (args: {
...
+        : files.components[name]
+      if (!component) {
+        // eslint-disable-next-line no-console
+        console.warn(`Unable to find component ${nodeLookupKey} in files`)
+        return undefined
+      }
+
+      return component
+    },
+    packageName: undefined,
```
erik-beus, line 70: The `packageName` is irrelevant for finding API dependencies within a single component.

jacob-kofoed (approved):
This is much better, no comments. I was not able to test as imports changed so need a branch on internal I suppose?

erik-beus:
> This is much better, no comments. I was not able to test as imports changed so need a branch on internal I suppose?

You can use this branch https://github.com/nordcraftengine/nordcraft-internal/pull/4390

jacob-kofoed (approved):

PR #1043, opened 2026-05-19
Issue rule for animated custom properties that are not in theme

jacob-kofoed:
Custom properties cannot be animated if they are not globally declared


erik-beus (approved):

erik-beus:
@jacob-kofoed is this ready to be merged? 🤞

PR #1044, opened 2026-05-19
Allow nullish values for more component properties

erik-beus:
- Adjust types for components to allow/support nullish values for attributes, variables, formulas, workflows, and events
- Filter out nullish values when iterating in all packages
- Adjust types for filterObject to take a second generic for the output type


jacob-kofoed (approved):
Tested and works. I hope we can make types strict in a future runtime version with some validation.

PR #1055, opened 2026-05-20
Remove legacy routing support and dependency

erik-beus:
Routing using the `page` property with a regex pattern has not been supported for several years now, and it should be safe to remove once we have verified that there's no traffic using that feature.

This reduces the runtime's size (page.main.js) from 32.5KB (114KB) to 30.2KB (108KB) which is a ~7% reduction


jacob-kofoed (approved):
I approve of this. We should test the release branch widely though.

erik-beus:
We still have a couple of projects using legacy routing. Instead of removing it completely, we should look into splitting the runtime in 2 (potentially more).

PR #1068, opened 2026-05-22
Include more CSS keywords for editor autocomplete

### LINKED ISSUE #1041: Can't add "auto" keyword into flex-basis field
According to [MDN](https://developer.mozilla.org/en-US/docs/Web/CSS/Reference/Properties/flex-basis) we should be able to use the "auto" keyword for flex-basis.

When trying to add "auto" into the flex basis field, it wants me to create a variable.

Image

It shows up correctly when I add "auto" via the CSS panel, though, but I can't add it via the style panel.

Image

Image

erik-beus:
- Handle `Property` type from `css-tree` to ensure we include additional CSS properties (e.g. `<'width'>` was not mapped to `auto`, `max-content`, `min-content` etc.)
- Write CSS keywords to a static file in the repo to make it easier to diff going forward (instead of building for every release)
- Adjust scripts

The actual changes to the CSS keywords can be found in [this commit diff](https://github.com/nordcraftengine/nordcraft/pull/1068/changes/7a3b9f06783244173041d8a7d06b46ef02031695#diff-d7b3602e623c11cdd223871dbb9eda33a6310df1a4218971ec167bff557f2958)
  
Closes https://github.com/nordcraftengine/nordcraft/issues/1041

From my localhost editor, auto completion for `flex-basis` looks like this:
CleanShot 2026-05-22 at 09 57 43


packages/editor/css-properties/cssPropertyKeywords.json
```diff
@@ -2208,119 +2346,82 @@
...
-    "border-bottom-color": [],
-    "border-bottom-left-radius": [],
-    "border-bottom-right-radius": [],
-    "border-bottom-style": [
+    "border-block-color": [],
+    "border-block-end": [
       {
-        "value": "none",
-        "description": "No border is drawn."
+        "value": "thin"
```
whitep4nth3r, line 2358: Looks like the descriptions for these values are missing?

packages/editor/css-properties/cssPropertyKeywords.json
```diff
@@ -2361,64 +2473,94 @@
...
       }
     ],
-    "border-left-color": [],
-    "border-left-style": [
+    "border-block-start-color": [],
+    "border-block-start-style": [
       {
-        "value": "none",
-        "description": "No border is drawn."
+        "value": "none"
```
whitep4nth3r, line 2485: Lots of descriptions missing here as well

whitep4nth3r (changes requested):
It looks like descriptions are missing from properties that used to have descriptions...

packages/editor/css-properties/cssPropertyKeywords.json
erik-beus: Previously, we did not have any values for `border-block-end`. See the actual diff here https://github.com/nordcraftengine/nordcraft/pull/1068/changes/7a3b9f06783244173041d8a7d06b46ef02031695#diff-d7b3602e623c11cdd223871dbb9eda33a6310df1a4218971ec167bff557f2958

I'll see if I can find descriptions for the new properties that were added somehow 🤞

whitep4nth3r (approved):

PR #1144, opened 2026-06-08
Improved text-editing in canvas

jacob-kofoed:
Completely redesigned how we edit text in canvas by actually editing in the iframe with contenteditable instead of faking it as today. There were a few issues that made this difficult, like proxying mouse-events through the canvas to select text.

This PR also:
- Moved the drag-and-drop logic to separate files along with the logic for text editing
- Made the auto-size canvas more stable
- Noticeably improved performance by bypassing the report formula logic entirely when not reporting.

Use editor branch `dev` to test these changes as the canvas needs to subscribe to the new update text event.

### DIFF
--- a/packages/core/src/formula/formula.ts
+++ b/packages/core/src/formula/formula.ts
@@ -209,13 +209,55 @@ export const isToddleFormula = <Handler>(
 
 export function applyFormula(
   formula: Formula | string | number | undefined | null | boolean,
-  _ctx: FormulaContext,
+  ctx: FormulaContext,
   extendedPath?: Array<string | number> | undefined,
 ): any {
-  const path = [...(_ctx?.jsonPath ?? []), ...(extendedPath ?? [])]
-  const ctx = { ..._ctx, jsonPath: path }
-  const report = (value: any, p: Array<string | number> = path) => {
-    ctx?.reportFormulaEvaluation?.(p, value, ctx)
+  // Short-circuit when not reporting to avoid unnecessary overhead of creating new objects and function
+  if (!ctx.reportFormulaEvaluation) {
+    if (!isFormula(formula)) {
+      return formula
+    }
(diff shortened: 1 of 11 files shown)


packages/core/src/formula/formula.ts
```diff
@@ -209,13 +209,55 @@ export const isToddleFormula = <Handler>(
...
+        case 'object':
+          return applyObjectFormula(formula, ctx)
+        case 'record':
+          return applyRecordFormula(formula, ctx)
+        case 'array':
+          return applyArrayFormula(formula, ctx)
+        case 'function':
+          return applyFunctionFormula(formula, ctx)
+        case 'apply':
+          return applyApplyFormula(formula, ctx)
```
marija04, line 241: Don't we need to include the `jsonPath` in the ctx we are having here and the functions above?

jacob-kofoed: I believe the `jsonPath` is only used by the reporter function (only runs in editor-preview), so if the reporter is not given, then the `jsonPath` is never used.

Let me know if you notice it being used anywhere else. This is a pretty large performance-hugger, so skipping all the reporter logic when not in reporting mode is pretty important if possible.

marija04 (approved):
I have tested it and seems to work fine.

jacob-kofoed:
This PR is ready, but awaiting me back from vacation to handle any potential issues. Note: Remember to release the `dev` editor branch along with this PR.

packages/ssr/src/rendering/testData.ts
```diff
@@ -199,7 +199,6 @@ const processNodes = (
             'style-variables',
             'style',
             'styleVariables',
-            'variants',
```
jacob-kofoed, line 202: This is currently breaking some builds.

PR #1146, opened 2026-06-08
Fix a duplicate ID issue

jacob-kofoed:
A duplicate ID could sometimes be created in repeat nodes by adding to the beginning of a list, with repeat-keys.


jacob-kofoed:
Can be tested at https://repeat-duplicate-id-jacob_test.nordcraft.site/home when changing search from "Do" to "D" and seeing that there are no duplicate Ids `0.1.0[default]`

marija04 (approved):
Works great 💪

PR #1151, opened 2026-06-15
Use the actual delay and duration

marija04:


packages/runtime/src/utils/getThemeSignal.ts
```diff
@@ -10,6 +10,16 @@ import { THEME_COOKIE_NAME } from '@nordcraft/core/dist/styling/theme.const'
 import { isDefined } from '@nordcraft/core/dist/utils/util'
 import { signal, type Signal } from '../signal/signal'
 
+export const clamp = (val: number, min: number, max: number) =>
```
jacob-kofoed: These two functions should probably live in some different util files, not really related to getThemeSignal

jacob-kofoed (approved):

PR #1165, opened 2026-06-19
Text editing improvements

jacob-kofoed:
Do not autofocus text nodes when selected.

### DIFF
--- a/packages/runtime/src/editor/drag-drop/dragMove.ts
+++ b/packages/runtime/src/editor/drag-drop/dragMove.ts
@@ -1,4 +1,5 @@
 import { findNearestLine } from '../../utils/findNearestLine'
+import { stripNodeIdRepeatIndices } from '../../utils/nodes'
 import type { DragState } from '../types'
 import { DRAG_REORDER_CLASSNAME } from './dragReorder'
 import { removeDropHighlight, setExternalDropHighlight } from './dropHighlight'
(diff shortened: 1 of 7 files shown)


packages/runtime/src/editor/drag-drop/dragMove.ts
```diff
@@ -78,7 +79,10 @@ export function dragMove(dragState: DragState | null, exclude: HTMLElement[]) {
     window.parent?.postMessage(
       {
         type: 'highlight',
-        highlightedNodeId: insertArea.parent.getAttribute('data-id'),
+        highlightedNodeId: stripNodeIdRepeatIndices(
+          insertArea.parent.getAttribute('data-id'),
```
erik-beus: Consider extracting this to a variable to avoid looking it up twice

packages/runtime/src/editor/drag-drop/dragReorder.ts
```diff
@@ -33,7 +34,11 @@ export function dragReorder(dragState: DragState | null) {
     window.parent?.postMessage(
       {
         type: 'highlight',
-        highlightedNodeId: dragState.initialContainer.getAttribute('data-id'),
+        highlightedNodeId: stripNodeIdRepeatIndices(
```
erik-beus: Same as above etc.

erik-beus (commented):
Please see suggestions

erik-beus (approved):

PR #1170, opened 2026-06-22
Omit empty style objects+fix classname consistency

erik-beus:
Earlier, we tried to strip empty style objects from the serialized data, but this introduced some styling issues. This happened since the client-side classname generation would generate classnames based on the stripped data, while the stylesheet generation (server side) would use the unstripped/raw data. This meant that classnames could differ between the server and client, which caused styles to not be applied correctly.
This PR ensures that classnames are consistent whether empty style objects/null/undefined are provided or not. It treats empty arrays, empty objects, null, and undefined as the same when generating classnames.

Later, we'll stop generating classnames client-side entirely, but this should fix the first issue that we saw earlier.

### DIFF
--- a/packages/core/src/styling/className.ts
+++ b/packages/core/src/styling/className.ts
@@ -1,9 +1,22 @@
+import type { NodeStyleModel } from '../component/component.types'
+import type { Nullable } from '../types'
 import { generateAlphabeticName, hash } from './hash'
+import type { StyleVariant } from './variantSelector'
 
 // Classnames are reused a lot, and JS hashing is expensive, so there is benefit in caching them in a native hashmap.
 const CLASSNAME_LOOKUP = new Map<string, string>()
-export const getClassName = (object: any) => {
-  const stringified = JSON.stringify(object)
+export const getClassName = (
+  object: Array<Nullable<NodeStyleModel> | Nullable<StyleVariant[]>>,
+) => {
+  const stringified = JSON.stringify(
+    object.filter(
+      (item) =>
+        item !== null && // Skip nullish values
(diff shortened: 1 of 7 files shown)


packages/core/src/styling/className.ts
```diff
@@ -1,9 +1,22 @@
+import type { NodeStyleModel } from '../component/component.types'
+import type { Nullable } from '../types'
 import { generateAlphabeticName, hash } from './hash'
+import type { StyleVariant } from './variantSelector'
 
 // Classnames are reused a lot, and JS hashing is expensive, so there is benefit in caching them in a native hashmap.
 const CLASSNAME_LOOKUP = new Map<string, string>()
-export const getClassName = (object: any) => {
-  const stringified = JSON.stringify(object)
+export const getClassName = (
```
jacob-kofoed, line 8: This is used on client as well, where we already strip away null-items with your changes, right? Maybe the logic to remove value should be added to the call to this function from SSR rather than in the shared function here?

Not a big deal, but if we clean types early, we should be able to rely on it in here. It also enables us to skip a few checks client side and save a few bytes on the runtime. But you expect to remove client side class-generation entirely?

jacob-kofoed (approved):
One question regarding where this logic should live.

packages/core/src/styling/className.ts
erik-beus: This PR makes part of this change redundant I suppose. But we still need to generate classnames client side in the editor's preview runtime https://github.com/nordcraftengine/nordcraft/pull/1018

PR #1179, opened 2026-06-25
Improve code reference detection performance

erik-beus:
- Return early if a component has custom action or project formula references
- Remove code from `ToddleComponent` that was only used in `codeRefs.ts`
- Add a few tests
- Adjust types


packages/ssr/src/custom-code/codeRefs.ts
```diff
@@ -140,11 +197,33 @@ export function takeReferencedFormulasAndActions({
...
 
 export const hasCustomCode = (component: Component, files: ProjectFiles) => {
-  const code = takeReferencedFormulasAndActions({ component, files })
-  return Object.values(code).some(
-    (c) =>
-      Object.keys(c.actions).length > 0 || Object.keys(c.formulas).length > 0,
-  )
+  const nordcraftComponent = componentClass(component, files)
+  const selfAction = nordcraftComponent.actionModelsInComponent().next().value
+  if (selfAction) {
```
jacob-kofoed, line 212: Why would this cause it to return true. Isn't it only if the action is pointing to a global action, not a lib-one?

erik-beus: Good point! Before, those actions were only filtered out because we are filtering project actions in the end by comparing whether they're part of the `Set` of `actionRefs`. I've updated the code now to filter out all `@toddle/` actions early on.

erik-beus: I've also updated it to ignore builtin formulas now 👍

jacob-kofoed (approved):
Looks good to me!

PR #1193, opened 2026-07-02
Save page state in memory for self-hosted backends

erik-beus:
- Avoid serializing the full page state in the HTML response, but delay it to a separate request
- This makes the HTML response smaller and faster to load
- It also causes less confusion for crawlers, since they don't need to parse the full page state in the HTML response
  
To test:
- Run the `backend` application using `bun run dev`
- Verify the page state is loaded in a separate request
- Verify custom code and logic on the page(s) work

Since I added [nanoid](https://npmx.dev/package/nanoid) as a dependency, I had to change the `esbuild` platform from `node --> browser` to ensure it doesn't use crypto from Node.js. This was solely to fix a compatibility issue with Cloudflare and to avoid adding the node compat flag in the wrangler file.


ChrEsb (approved):

packages/backend/src/routes/nordcraftPage.ts
```diff
@@ -220,6 +185,29 @@ export const nordcraftPage = async ({
...
+  // Persist the page state in a cache so that it can be retrieved by the client for hydration
+  const { pageStateUrl, customCodeUrl } = cachePageState({
+    state: toddleInternals,
+    ctx: hono,
+    pageName: page.name,
+    usesCustomCode,
+    options,
+  })
+  const additionalHeadScripts = [
+    `<link rel="modulepreload" href="/_static/page.main.esm.js")}</link>`,
```
max-kayr: @erik-beus the `link` tags don't need a closing tag

erik-beus: Fixed 👍

PR #1195, opened 2026-07-03
Add 2 issue rules to flag invalid path formulas

### LINKED ISSUE #1158: Missing issue rule
The AI generated this thing, but there was no issue rule to pick it up.

Image

````
{"type":"formula","formula":{"type":"path","path":["Formulas","isClearLabel"]}}
```

Formulas are applied, so it is not valid to use a path formula

erik-beus:
This PR introduces 2 new rules:
- 1 that detects invalid path formulas to formulas/workflows
- 1 that detects invalid references to workflow parameters in formulas that are not part of a workflow

Closes https://github.com/nordcraftengine/nordcraft/issues/1158

### DIFF
--- a/packages/search/src/rules/issues/formulas/workflowParameterOutsideWorkflowRule.ts
+++ b/packages/search/src/rules/issues/formulas/workflowParameterOutsideWorkflowRule.ts
@@ -1,0 +1,23 @@
+import type { IssueRule } from '../../../types'
+import { isPathFormula } from '../../../util/formulas'
+
+export const workflowParameterOutsideWorkflowRule: IssueRule = {
+  code: 'invalid path formula',
+  level: 'error',
+  category: 'Unknown Reference',
+  visit: (report, { path, value, nodeType }) => {
+    if (nodeType !== 'formula' || !isPathFormula(value)) {
+      return
+    }
+
+    if (value.path.at(0) === 'Parameters' && !path.includes('workflows')) {
+      report({
+        path,
+        info: {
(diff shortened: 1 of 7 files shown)


packages/search/src/rules/issues/formulas/workflowParameterOutsideWorkflowRule.ts
```diff
@@ -0,0 +1,23 @@
...
+export const workflowParameterOutsideWorkflowRule: IssueRule = {
+  code: 'invalid path formula',
+  level: 'error',
+  category: 'Unknown Reference',
+  visit: (report, { path, value, nodeType }) => {
+    if (nodeType !== 'formula' || !isPathFormula(value)) {
+      return
+    }
+
+    if (value.path.at(0) === 'Parameters' && !path.includes('workflows')) {
```
jacob-kofoed, line 13: Should we check if the parameters exist on the given workflow? Perhaps a separate rule for this is better.

jacob-kofoed (approved):

packages/search/src/rules/issues/formulas/workflowParameterOutsideWorkflowRule.ts
erik-beus: We already have a rule that checks for invalid workflow parameter names [here](https://github.com/nordcraftengine/nordcraft/blob/main/packages/search/src/rules/issues/workflows/unknownWorkflowParameterRule.ts). Isn't that what you meant?

PR #1197, opened 2026-07-07
Improve context update handling in editor preview

erik-beus:
Ensure context providers are updated when components are changed. This prevents stale context where a component might not be able to read a value correctly from a context provider's formula if the formula was updated.


jacob-kofoed (approved):
Seems to work for me

PR #1198, opened 2026-07-07
Fix an issue where duplicate IDs with repeating slots could happen

jacob-kofoed:


packages/runtime/src/editor/style.ts
```diff
@@ -264,7 +264,9 @@ export const styleToCss = (style: NodeStyleModel) => {
 export const convertViewportUnitsToEmulatedViewportUnits = (
```
erik-beus, line 264: Perhaps this function should be moved into the `editor/` folder to ensure it doesn't add unnecessary bytes to the actual runtime?

erik-beus (approved):

PR #1199, opened 2026-07-07
Add issue rules for unnecessary name properties

erik-beus:
The `name` property for component formulas and workflows are no longer necessary and should be removed throughout Nordcraft projects. This PR adds rules to detect and report these unnecessary properties, as well as fixes to remove them when found.

Getting rid of the `name` property will:
- Improve the AI's ability to work with projects
- Make it easier to read the raw project structure
- Avoid multiple sources of truth
- Make it easier to understand broken references

### DIFF
--- a/packages/search/src/rules/issues/workflows/namedComponentWorkflowRule.ts
+++ b/packages/search/src/rules/issues/workflows/namedComponentWorkflowRule.ts
@@ -7,7 +7,7 @@ export const namedComponentWorkflowRule: IssueRule<
 > = {
   code: 'named component workflow',
   level: 'info',
-  category: 'Quality',
+  category: 'Deprecation',
   visit: (report, { nodeType, value, path }) => {
     const workflowKey = path.at(-1)
     if (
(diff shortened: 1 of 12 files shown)


packages/search/src/rules/issues/workflows/namedComponentWorkflowRule.ts
```diff
@@ -0,0 +1,40 @@
+import type { ComponentWorkflowNode, IssueRule } from '../../../types'
+import { renameNamedComponentWorkflowFix } from './namedComponentWorkflowRule.fix'
+
+export const namedComponentWorkflowRule: IssueRule<
+  { name: string; workflowKey: string | number },
+  ComponentWorkflowNode
+> = {
+  code: 'named component workflow',
+  level: 'info',
+  category: 'Quality',
```
jacob-kofoed: Maybe category "Deprecation" is more correct?

packages/search/src/rules/issues/workflows/namedComponentWorkflowRule.ts
```diff
@@ -0,0 +1,40 @@
...
+      typeof workflowKey !== 'string'
+    ) {
+      return
+    }
+
+    report({
+      path,
+      info: {
+        title: 'Workflow has name property',
+        description: `The name for the **${value.name}** workflow is declared directly on the workflow. Consider moving the name to the workflow key to reduce complexity in your project and help the AI perform better.`,
```
jacob-kofoed, line 26: Maybe frame this more like a deprecation description. The description is very internal and may not make sense to most users unless they've inspected the format. I suggest something like:

`Workflow **${value.name}* has a legacy name field. Consider migrating the workflow to use standard keys by using the "..." fix.`

jacob-kofoed (approved):

PR #1227, opened 2026-07-14
CSS variables on variants are now part of class styling

jacob-kofoed:
This fixes an issue with dynamic custom properties after we removed style and variant data during preprocessing.

This PR also improves general performance as class selectors are faster than attribute selectors – as a note on this, we should also modify the dynamic properties to target a class rather than attribute selectors. This is a requirement to get rid of noisy `data-id` & `data-node-id` attributes everywhere.

### DIFF
--- a/packages/ssr/src/rendering/components.test.ts
+++ b/packages/ssr/src/rendering/components.test.ts
@@ -360,7 +360,7 @@ describe('renderPageBody', () => {
     // Expect the nested child text node inside the package component to also render the formula value,
     // proving that the formula is evaluated correctly even when used inside a package component
     expect(html).toInclude(
-      '<span data-id="0.1" data-node-id="root" class="PKfF PageComponent:secondChild"><span data-node-type="text" data-node-id="nestedChild">output-from-test-formula</span></span>',
+      '<span data-id="0.1" data-node-id="root" class="PageComponent:secondChild"><span data-node-type="text" data-node-id="nestedChild">output-from-test-formula</span></span>',
     )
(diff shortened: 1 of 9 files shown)


packages/ssr/src/rendering/components.test.ts
```diff
@@ -360,7 +360,7 @@ describe('renderPageBody', () => {
     // Expect the nested child text node inside the package component to also render the formula value,
     // proving that the formula is evaluated correctly even when used inside a package component
     expect(html).toInclude(
-      '<span data-id="0.1" data-node-id="root" class="PKfF PageComponent:secondChild"><span data-node-type="text" data-node-id="nestedChild">output-from-test-formula</span></span>',
```
jacob-kofoed, line 363: This was a bit silly, but empty classes (no styles at all) would hash to `"PKfF"`, but no need to send this info.

packages/ssr/src/rendering/components.test.ts
```diff
@@ -961,7 +961,7 @@ describe('renderPageBody', () => {
 
     // Expect the ID formula to render with the correct order (0 in this case since it's the first formula rendered)
     expect(html).toBe(
-      `<div data-custom-id="custom_id_0_" data-id="0" data-node-id="root" class="PKfF"></div>`,
```
jacob-kofoed, line 964: Same here, now we even strip the class entirely to save a few more bytes

marija04 (approved):
Works good ✅

PR #1228, opened 2026-07-14
Better parsing of the variables used in the style

marija04:
Fix the parser so we better support variables that reference another variables.


jacob-kofoed (approved):
cleaner

PR #1236, opened 2026-07-15
Better custom element preprocessing of classes and style variables

jacob-kofoed:
Fixes an issue where the class hashes for custom elements would not always match the generated classes.

This require some changes to internal, similar to `routes/customElement.ts`.

### DIFF
--- a/packages/ssr/src/rendering/classes.ts
+++ b/packages/ssr/src/rendering/classes.ts
@@ -13,51 +13,67 @@ import { mapObject } from '@nordcraft/core/dist/utils/collections'
 
 /**
  * Function to strip styles and variants from a Component's nodes and convert them to static class names
+ *
+ * @param options - Options to control the behavior of the function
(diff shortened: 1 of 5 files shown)


packages/ssr/src/rendering/classes.ts
```diff
@@ -13,51 +13,67 @@ import { mapObject } from '@nordcraft/core/dist/utils/collections'
...
+                  : node.classes,
+            }),
+            style: options.clearStyle ? undefined : nodeStyle,
+            customProperties: filterDynamicProperties(node.customProperties),
+            variants:
+              hasVariantCustomProperties || !options.clearStyle
+                ? variants
+                : undefined,
+          },
+        ] as [string, NodeModel]
```
erik-beus: It would be great if you could avoid this type cast. It's usually better to create a variable of type NodeModel and return that

jacob-kofoed: The type-cast was not necessary at all. I suppose it was a leftover from when I was implementing - removed.

PR #1238, opened 2026-07-16
Use an internal path instead of reading on [data-id]

jacob-kofoed:
Work in progress. Consider merging after/with #1216 

This PR removes the `[data-id]` attribute from all rendered elements (when not in editor preview runtime – it is possible to remove it from editor runtime as well, but will require a bit more work).

This PR also change structure we store node paths internally by using a fixed structure rather than a query-string. This should have a positive impact on performance as repeat-node reordering has to do some string manipulation today.

## Note on `[data-node-id]`

It is very much possible to remove this attribute along with `[data-node-type]` & `[data-component-name]` from production output as well, but the reset stylesheet does require them today. By modifying (let users opt-out) of the reset stylesheet we are down to zero built-in attribute and render only what you see in the editor.

### DIFF
--- a/packages/runtime/src/utils/nodes.ts
+++ b/packages/runtime/src/utils/nodes.ts
@@ -1,8 +1,12 @@
+/* eslint-disable no-console */
 import type {
   Component,
   NodeModel,
 } from '@nordcraft/core/dist/component/component.types'
+import { pathToString, stringToPath } from '@nordcraft/core/dist/utils/path'
 import { isDefined } from '@nordcraft/core/dist/utils/util'
+import { PATH } from '../constants'
+import type { Path } from '../types'
 
 type NodeWithNodeId = NodeModel & { nodeId: string }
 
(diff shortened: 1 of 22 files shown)


packages/runtime/src/utils/nodes.ts
```diff
@@ -125,7 +129,16 @@ export function ensureEfficientOrdering(
...
       // The item is either not in the DOM or not in the correct position.
       // Insert the item before the insertBeforeElement (or append it if insertBeforeElement is null).
-      parentElement.insertBefore(item, insertBeforeElement)
+      if (
+        item.parentNode === parentElement &&
+        'moveBefore' in Element.prototype
+      ) {
+        // `moveBefore` is not yet supported in Safari and some older browsers,
+        // moveBefore actually moves the element in the DOM instead of removing and reinserting it, which is more efficient and preserves animation-, transition- and focus states.
+        parentElement.moveBefore(item, insertBeforeElement)
```
jacob-kofoed, line 138: This is actually a pretty exciting change. It means animations can now keep playing from where they are at when using a repeat-key. Input elements will keep their focus and text-ranges selected when moved around as well. It should also be more performant overall as teardown & construction can be skipped.

PR #1241, opened 2026-07-24
Allow the iframe to take a screenshot on demand

ChrEsb:
This PR introduces a new message type to the iframe. The message requests the iframe to generate a "screenshot" of its content. It then returns this content to the caller.


packages/runtime/src/editor-preview.main.ts
```diff
@@ -1148,6 +1149,39 @@ body[data-mode="design"] [data-id="${animationState.animatedElementId}"], body[d
...
+        case 'capture_screenshot': {
+          const { id } = message.data
+          try {
+            // Rasterize the full document (not just what's currently scrolled into
+            // view) by walking the DOM/computed styles rather than capturing the
+            // screen, so content below the fold is still included.
+            const target = document.documentElement
+            const canvas = await domToCanvas(target, {
+              width: target.scrollWidth,
+              height: target.scrollHeight,
```
jacob-kofoed, line 1199: What is height used for if everything below the fold is still included? Just curious how this works.

jacob-kofoed (approved):

packages/runtime/src/editor-preview.main.ts
jacob-kofoed: Also, since the image is generated independently, maybe the AI wants to have an argument to set the viewport width? To test certain breakpoints?

ChrEsb: Thats a good idea.

Regarding height/width, its because the domToCanvas expects it. Otherwise it falls back to trying to calculate what is currently in view. At least I think so. Im testing it out currently

ChrEsb: > Also, since the image is generated independently, maybe the AI wants to have an argument to set the viewport width? To test certain breakpoints?

Actually Im not sure this is possible as we would need to set this on the iframe itself rather than the DOM inside it. So instead I will try to make this fix in the editor.

jacob-kofoed (approved):

PR #1255, opened 2026-08-04
Use fast-equals instead of fast-deep-equal

jacob-kofoed:
Waiting for performance-metrics before this is to be merged.

According to standard benchmarks this should give a 1.2x improvement to equality checks which is often responsible for up to 50% of time taken for state changes. So this should give an overall **1.1x** improvement to performance, but needs to be validated first.

### DIFF
--- a/bun.lock
+++ b/bun.lock
@@ -68,7 +68,6 @@
       "name": "@nordcraft/std-lib",
       "dependencies": {
         "@nordcraft/core": "workspace:*",
-        "fast-deep-equal": "3.1.3",
         "fast-equals": "6.0.2",
       },
       "devDependencies": {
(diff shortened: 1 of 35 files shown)


erik-beus:
@jacob-kofoed this is very interesting 👀
It does seem like the fast-equals package is a somewhat larger dependency though? https://bundlephobia.com/package/fast-equals vs https://bundlephobia.com/package/fast-deep-equal
It's probably not enough that it matters but worth keeping in mind

jacob-kofoed:
> @jacob-kofoed this is very interesting 👀 It does seem like the fast-equals package is a somewhat larger dependency though? https://bundlephobia.com/package/fast-equals vs https://bundlephobia.com/package/fast-deep-equal It's probably not enough that it matters but worth keeping in mind

I noticed this as well. If it actually works and is faster, I suggest we take just the parts that we need and add as code to the project. It seems to have a lot of advanced option flags that we don't need.

PR #1256, opened 2026-08-05
Perf: Improved collection helper functions

jacob-kofoed:
I ran each function through benchmarks some 100.000 iterations with warmup runs beforehand. Improvements for each function ranges from ~1.3x to ~4x in separation. Actual improvement for SSR or runtime setup and handling changes was not profiled yet.

This PR is a good candidate to test some automatic performance benchmarking @erik-beus as well as bundle-size increase/decrease as this adds more lines of code, but may be able to compress more than the previous code.

### DIFF
--- a/packages/core/src/utils/collections.ts
+++ b/packages/core/src/utils/collections.ts
@@ -10,8 +10,9 @@ export const mapObject = <T, T2>(
 ): Record<string, T2> => {
   const result: Record<string, T2> = {}
   for (const key in object) {
-    const [k, v] = f([key, object[key] as T])
-    result[k] = v
+    const v = object[key] as T
+    const [k, mappedV] = f([key, v])
+    result[k] = mappedV
   }
   return result
 }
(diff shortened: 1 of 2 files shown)


packages/core/src/utils/collections.ts
```diff
@@ -10,16 +10,23 @@ export const mapObject = <T, T2>(
...
   }
   return result
 }
 
 export const mapValues = <T, T2>(
   object: Record<string, T>,
   f: (value: T) => T2,
-): Record<string, T2> => mapObject(object, ([key, value]) => [key, f(value)])
+): Record<string, T2> => {
+  const result = {} as Record<string, T2>
```
erik-beus:
```suggestion
  const result: Record<string, T2> = {}
```

packages/core/src/utils/collections.ts
```diff
@@ -67,51 +80,76 @@ export const omitPaths = <T extends Record<string, any>>(
...
 
 export const filterObject = <T, T2 extends T = T>(
   object: Record<string, T>,
   f: (kv: [string, T]) => boolean,
-): Record<string, T2> =>
-  Object.fromEntries(
-    Object.entries(object).filter((o): o is [string, T2] => f(o)),
-  )
+): Record<string, T2> => {
+  const result = {} as Record<string, T2>
```
erik-beus:
```suggestion
  const result: Record<string, T2> = {}
```

erik-beus (commented):
Looks good 💪 2 small suggestions
Also, I agree that this would be fun to benchmark 🙏

PR #1263, opened 2026-08-07
Text rects more precise

jacob-kofoed:
Text rects would not always fully surround the text, which is important for in-canvas editing of text-nodes. The logic is increasingly complex to handle what the browser does (though no API for it).

This PR:
- Also improve performance of className validation which is used a lot both client and server side.
- Replaced BatchQueue implementation with just queueMicrotask as is it technically the same after we moved from requestAnimationFrame to microtask. This should save a few bytes on bundle-size and have no negative consequences on performance (though a benchmark on this would be nice).


packages/runtime/src/utils/BatchQueue.ts
```diff
@@ -1,22 +1,36 @@
-/**
- * A helper class to batch multiple callbacks and process them in a single update step just before the next frame render, but after the current stack.
- * This is more efficient than processing each callback in a separate requestAnimationFrame due to the overhead.
- */
 export class BatchQueue {
```
jacob-kofoed, line 1: Made this implementation slightly more capable. The tasks added to the queue here are not required to run immediately, so it should be safe to postpone tasks to next macrotask. I put in a 5ms batch time to allow us to paint in between tasks. This should improve perceived performance for heavy updates and first-render.

A schedular would probably be more better here as order of execution is not really important, but we only use this class very limited at the moment, so no obvious gain? It would be interesting to look into https://developer.mozilla.org/en-US/docs/Web/API/Scheduler when/if baseline and yield tasks on any user-input.

packages/core/src/styling/className.ts
```diff
@@ -92,8 +92,11 @@ export const toValidClassName = (
   }
 
   // Ensure the class name doesn't start with a number or special character
-  if (/^[^a-zA-Z]/.test(className)) {
-    className = `_${className}`
+  if (className.length > 0) {
```
jacob-kofoed, line 95: Same as before, but avoids the regex. We generally should avoid any regex in the runtime and SSR code as they are major performance pitfalls.

packages/runtime/src/components/createElement.ts
```diff
@@ -76,10 +77,11 @@ export function createElement({
   // class names are baked during preprocessing, except for in editor-preview where we generate them on the fly
   if (node.style || node.variants?.some((v) => v.style)) {
     const classHash = getClassName([node.style, node.variants])
-    elem.classList.add(classHash)
+    initialClasses.push(classHash)
```
jacob-kofoed, line 80: Using a temporary list should bundle to a smaller size and is also faster performance-wise to add multiple classes in one go vs. one at a time.

packages/runtime/src/editor/overlay.ts
```diff
@@ -27,73 +27,104 @@ export function getRectData(selectedNode: Element | null | undefined) {
...
+  if (matrix.isIdentity) {
+    return rect
   }
 
-  nodeEl.style.setProperty(
-    'transform',
-    `${new DOMMatrix(rotate).inverse().toString()} ${baseTransform}`,
-    'important',
-  )
+  const inverse = matrix.inverse()
```
jacob-kofoed, line 43: Very ugly helper file to get the actual overlay rect for nodes no matter their matrix or any inherited matrix multiplication from ancestors.

packages/runtime/src/editor-preview.main.ts
```diff
@@ -562,7 +563,11 @@ export const createRoot = (
               node.getAttribute('data-node-type') === 'text'
             ) {
               requestAnimationFrame(() => {
-                handleTextNodeSelection(node)
+                handleTextNodeSelection(node, {
+                  onInput: () => {
+                    syncOverlayRects()
```
jacob-kofoed, line 568: This should cause the overlay rect to update whenever the user types while editing a text-node. It did not always work before.

packages/runtime/src/editor-preview.main.ts
```diff
@@ -881,10 +889,10 @@ export const createRoot = (
                   .map((value) => {
                     // If it is a float or float with unit we want to round to 2 decimal
                     if (value.match(/^(-?\d+)\.\d+([a-z]*|%?)$/)) {
-                      const splited = value.match(/([0-9.]+)\s*(.*)/) ?? ''
```
jacob-kofoed, line 884: typo*

packages/runtime/src/editor-preview.main.ts
```diff
@@ -1132,8 +1140,17 @@ body[data-mode="design"] [data-id="${animationState.animatedElementId}"], body[d
               resourceElement.rel = 'stylesheet'
               resourceElement.href = resource.href
               document.head.appendChild(resourceElement)
+
+              // Sync canvas after the resource has loaded (if not already loaded)
+              if (!resourceElement.sheet) {
```
jacob-kofoed, line 1145: Not an actual bug I witnessed, but in case a font loads and it pushes content around, the current selection/highlight rects may have moved 🤔

marija04 (approved):
✅

PR #1284, opened 2026-08-21
Highlight position on insert

marija04:
You can test it, using this branch 👇
https://editor.nordcraft.com/projects/toddle/branches/nc-for-designers/components/EditorPage?canvas-width=800&rightpanel=events


packages/runtime/src/editor-preview.main.ts
```diff
@@ -855,6 +872,23 @@ export const createRoot = (
             )
           }
           break
+        // Not sure if I need tthis ?
```
jacob-kofoed: Probably need this if we have hotkeys for inserting? I noticed we have an issue today when pressing `[meta-key]` for trying to target text-nodes, that it only works after moving the cursor, but it should update immediately.

packages/runtime/src/editor/insert/insertEnded.ts
```diff
@@ -0,0 +1,82 @@
+import { isElementInViewport } from '../../utils/isElementInViewport'
+import { tryStartViewTransition } from '../../utils/tryStartViewTransition'
+import { removeDropHighlight } from '../drag-drop/dropHighlight'
+import type { InsertState } from '../types'
+import { INSERT_MOVE_CLASSNAME } from './insertMove'
+import { INSERT_REORDER_CLASSNAME } from './insertStarted'
+
+export async function insertEnded(insertState: InsertState, canceled: boolean) {
```
jacob-kofoed: Are these copy/paste of the drag/drop? I suppose we should refactor to combine at least the line-rendering logic and probably more.

marija04: Yes, there might be some changes but it is mostly the same. So I'm planning to refactor a bit so we can use it in the both places.

packages/runtime/src/editor/insert/insertHandlers.ts
```diff
@@ -0,0 +1,70 @@
...
+
+export const handleInsertEnded = async (
+  messageData: { canceled?: boolean },
+  insertState: DragInsertState,
+): Promise<DragInsertState | null> => {
+  const selectedPermutation =
+    insertState?.insertAreas?.[insertState?.selectedInsertAreaIndex ?? -1]
+  if (selectedPermutation && !messageData.canceled) {
+    await dragInsertEnded(insertState, false)
+
```
marija04, line 70: @jacob-kofoed instead of sending the message to the editor and then creating the element there, is there a way to create it from here and if yes should I do that?

jacob-kofoed: There is a noticeable delay if you wait for the editor to update after modifying files. I suggest you first add it here so it is there immediately and then also update the editor. As the files updates it should be exactly the same so things happens instantly from the user's perspective.

We do it similar with drag & drop today where we move the element optimistically and then let the standard files update handle the real move afterwards.

jacob-kofoed (approved):
I tested and it works well. I think it is possible for both drag-and-drop and the new insert to allow inserting items into elements with no child-nodes. In this branch, inserting a div and then trying to insert a text-node into it is not possible.

Maybe we can add an insert line at centre or something for elements with no child-nodes (but not for self-closing tags like IMG etc.)

jacob-kofoed (approved):
I tested this and it works well. Inserting nodes into empty nodes adds a line at the top of the element. It kinda makes sense but it often overlaps with a line just outside the node. Perhaps the insert areas of empty nodes should be in their centre instead of at the start.

PR #1291, opened 2026-08-28
Bump version

marija04:


packages/runtime/src/editor/drag-drop/getInsertAreas.ts
```diff
@@ -38,7 +38,7 @@ export function getInsertAreas() {
           index: 0,
           center: {
             x: rect.left + rect.width / 2,
-            y: rect.top,
```
marija04, line 41: I forgot to add this change in my previous PR.

jacob-kofoed (approved):

PR #1293, opened 2026-08-31
Add SSR benchmark scripts + workflow

erik-beus:


jacob-kofoed (approved):
Very cool!

You can try adding https://bun.com/reference/bun/gc (true) just before the benchmark actually starts to not spend time on things outside the test-case itself.

Besides that more warmup and iterations may improve stability, but I think we can start with this and then see how it performs and improve/add benchmarks as we go.

PR #1307, opened 2026-09-04
Bump Bun version to 1.4.1

erik-beus:
See release notes here https://bun.com/blog/bun-v1.4.1
This should help speed up self hosted (Docker) builds significantly.

### DIFF
--- a/package.json
+++ b/package.json
@@ -15,7 +15,7 @@
     "@html-eslint/eslint-plugin": "0.65.0",
     "@html-eslint/parser": "0.65.0",
     "@octokit/core": "7.0.6",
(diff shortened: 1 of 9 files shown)


package.json
```diff
@@ -15,7 +15,7 @@
     "@html-eslint/eslint-plugin": "0.65.0",
     "@html-eslint/parser": "0.65.0",
     "@octokit/core": "7.0.6",
-    "@types/bun": "1.3.14",
+    "@types/bun": "1.4.0",
```
erik-beus, line 18: Note that types for bun 1.4.1 are not available (yet)

jacob-kofoed (approved):

PR #1308, opened 2026-09-04
Overlay supports skew and is overall more precise

jacob-kofoed:
With addition of more complex in-canvas editing tools it is even more important that the overlay div matches the inner elements. This PR improves precision and supports transform skew.


marija04 (approved):
Seems to work fine 🙌

PR #1311, opened 2026-09-07
Bump bun to 1.4.2

erik-beus:
See release notes here https://bun.com/blog/bun-v1.4.2

This also bumps the types for bun to 1.4.2 (latest)


jacob-kofoed (approved):
Works for me ✅

PR #1327, opened 2026-09-21
Ensure signal subscriptions are properly managed

erik-beus:
- Unsubscribe from global theme store when a component is destroyed
- Unsubscribe from context subscriptions when a component is destroyed


jacob-kofoed (commented):
I was able to test this to make sure that we do not have any signal leaks. With your changes I couldn't see any, but also without your changes it seemed to be cleared up. Are we sure the data was not cleaned up before automatically? Subscribing or mapping from one signal should automatically clean up derived signals afaik.

erik-beus:
> I was able to test this to make sure that we do not have any signal leaks. With your changes I couldn't see any, but also without your changes it seemed to be cleared up. Are we sure the data was not cleaned up before automatically? Subscribing or mapping from one signal should automatically clean up derived signals afaik.

You could be right. I'll check on `main` and compare 👍 
Some of the adjustments in this PR might still be relevant, but let's see when we know for sure if it solves a real issue or not

PR #1330, opened 2026-09-21
Pass data as separate argument in formulas

jacob-kofoed:
Work in progress to improve performance and reduce memory usage by constructing fewer objects during runtime. This will require an update to nordcraft-internal and potentially need a major release bump unless it is made backwards compatible.


packages/backend/src/utils/api.ts
```diff
@@ -122,31 +126,25 @@ const fetchApi = async ({
...
-  }
-
-  const newFormulaContext: FormulaContext = {
-    component: formulaContext.component,
-    package: formulaContext.package,
-    data,
-    env: formulaContext.env,
-    toddle: formulaContext.toddle,
-  }
+  } as ComponentData
```
erik-beus, line 138: This typecast should not be needed?

```suggestion
  }
```

packages/core/src/formula/formula.ts
```diff
@@ -141,7 +141,7 @@ export type Formula =
   | ApplyOperation
 
 export interface FormulaContext {
-  component: Component | undefined
+  component?: Component | undefined
```
erik-beus, line 144: I kind of liked not making this optional since it ensured we only left it out on purpose

packages/core/src/formula/formula.ts
```diff
@@ -151,9 +151,8 @@ export interface FormulaContext {
       }
     >
   >
-  data: ComponentData
   root?: Nullable<Document | ShadowRoot>
-  package: Nullable<string>
+  package?: Nullable<string>
```
erik-beus, line 155: Same as above

packages/core/src/formula/formula.ts
```diff
@@ -211,10 +210,33 @@ export const isToddleFormula = <Handler>(
...
+export function applyFormula(
+  formula: Formula | string | number | undefined | null | boolean,
+  ctx: FormulaContext,
+  dataOrPath?: ComponentData | Array<string | number>,
   extendedPath?: Array<string | number> | undefined,
 ): any {
+  let resolvedData: ComponentData | undefined
+  let resolvedPath: Array<string | number> | undefined
+
+  if (Array.isArray(dataOrPath)) {
```
erik-beus, line 230: Would it be terrible to only have 1 implementation (no overrides) to avoid a call to `Array.isArray` for every formula evaluation?

packages/core/src/formula/formula.ts
```diff
@@ -211,10 +210,33 @@ export const isToddleFormula = <Handler>(
...
+    resolvedPath = dataOrPath
+    resolvedData = undefined
+  } else {
+    resolvedData = dataOrPath
+    resolvedPath = extendedPath
+  }
+
   // Short-circuit when not reporting to avoid unnecessary overhead of creating new objects and function
-  if (!ctx.reportFormulaEvaluation) {
+  if (!ctx?.reportFormulaEvaluation) {
```
erik-beus, line 239: When is `ctx` not defined?

packages/core/src/formula/formula.ts
```diff
@@ -223,30 +245,30 @@ export function applyFormula(
...
+          return applyApplyFormula(formula, ctx, resolvedData)
         default:
-          if (ctx.env?.logErrors) {
+          if (ctx?.env?.logErrors) {
             console.error('Could not recognize formula', formula)
           }
       }
     } catch (e) {
-      if (ctx.env?.logErrors) {
+      if (ctx?.env?.logErrors) {
```
erik-beus, line 271: Same question as above

packages/core/src/formula/functionFormula.ts
```diff
@@ -22,39 +25,49 @@ export const applyFunctionFormula = (
     ctx.toddle ??
     ((globalThis as any).toddle as Toddle<unknown, unknown> | undefined)
   )?.getCustomFormula(formula.name, packageName)
+  const resolvedData: ComponentData = data ?? { Attributes: {} }
```
erik-beus, line 28: Why is `Attributes` required in `ComponentData`? 🤔

erik-beus (changes requested):
Reviewed 12/44 files so far. I'll continue later

PR #1332, opened 2026-09-22
Update the insert areas used for insert and drag and drop

marija04:
This should allow the user to insert or move an element with drag/drop more precisely. For example before or after the element in the component slot.
Also now should be possible to move/insert an element between text that is not wrapped. 

Verify that when you are in insert mode the insert line is removed if you press escape.

To test it use this branch in the toddle project `remove-highlight-on-esc` and verify everything works fine.

Also


jacob-kofoed (approved):
Looks good. It would be nice if we could allow drag/drop of text-nodes in the editor after this PR is merged.

marija04:
> Looks good. It would be nice if we could allow drag/drop of text-nodes in the editor after this PR is merged.

Yes, I can have a look into that :)

PR #1333, opened 2026-09-22
Add runtime benchmark workflow and related files

erik-beus:
- Add Playwright support for runtime benchmarking
- Add fixtures for different projects
- Run benchmarks on commit to compare and check for regressions/improvements


jacob-kofoed (approved):
plz make output text more readable, the text wraps too much. Maybe combine base/head. Code looks good, thanks for explanation.

